PR #4, opened 2021-10-05
The GEVP method now has the parameter "state", deciding which eigenve…

JanNeuendorf:
…ctor to return.

Correlators get theoptional attribute "range", which is preserved under certain operations.
Fits and plateaus will default to this range.

### DIFF
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -3,6 +3,7 @@ import autograd.numpy as anp
 #from scipy.special.orthogonal import _IntegerType
 from .pyerrors import *
 from .fits import standard_fit
+from .linalg import *
 from .roots import find_root
 from matplotlib import pyplot as plt
 from matplotlib.ticker import NullFormatter  # useful for `logit` scale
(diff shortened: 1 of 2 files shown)


fjosw (changes requested):
Thanks for the contribution! As `range` is a python built-in function I would suggest to use a slightly different name.
I would also suggest to initialize the range to `None` in case no range is given. This would alter line 63 & 64 to
```
self.range = range
```
Consecutive initiations can then by called via
```
...
return Corr(newcontent, range=self.range)
```
And a range can be easily removed via
```
my_corr.range = None
```

JanNeuendorf:
Yes, i noticed, that "range is the name of a build in function. This is not actually a problem, since instance.attribute has a separate namespace, but i think you are right, since it might be confusing. I will change it to prange. 

As for the default being None, i decided against it for a reason. 
Since there is no archiving solution yet, dumping and loading Corrs is how i manage my data. 
That means, i have to be able to load data, which has been created before the Corr class was changed. 
The safest way to do that is to allow the Corr to simply not have the new attribute. 
While this reason is rather selfish, i think it is still a good idea. The (p)range is only one small addition, 
but we already talked about tags and there might be other more complicated attributes we will think of. 

One thing that you should keep a look at is the use of anp and np in the module. 
Should i change all the np calls to anp?

PR #8, opened 2021-11-15
Feature/corrfits

s-kuberski:
Add the possibility to constrain the last N fit parameters according to a list of Obs to least_squares and total_least_squares. 

Add the possibility to perform a correlated fit with least_squares. Currently only available with the standard method and without priors. A basic check of the stability of the inversion of the covariance matrix is implemented, however, this does not suffice to ensure the correctness of correlated fits. Might be improvable, but the basic functionality works. USE WITH CARE!  

Return 0. in covariance and covariance2 for the special case, where the two observables are defined on disjunct sets of ensembles.

### DIFF
--- a/pyerrors/fits.py
+++ b/pyerrors/fits.py
@@ -109,12 +109,11 @@ def least_squares(x, y, func, priors=None, silent=False, **kwargs):
         corrected by effects caused by correlated input data.
         This can take a while as the full correlation matrix
         has to be calculated (default False).
-    correlated_fit : int
+    correlated_fit : bool
         If true, use the full correlation matrix in the definition of the chisquare 
         (only works for prior==None and when no method is given, at the moment).
     const_par : list, optional
-        List of N Obs that are used to constrain the last N fit parameters of func and
-        to take into account the correlations.
+        List of N Obs that are used to constrain the last N fit parameters of func.
     '''
     if priors is not None:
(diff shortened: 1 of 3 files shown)


pyerrors/fits.py
```diff
@@ -507,32 +584,44 @@ def chisqfunc_residuals(p):
         print('chisquare/d.o.f.:', output.chisquare_by_dof)
 
     if kwargs.get('expected_chisquare') is True:
-        W = np.diag(1 / np.asarray(dy_f))
-        cov = covariance_matrix(y)
-        A = W @ jacobian(func)(fit_result.x, x)
-        P_phi = A @ np.linalg.inv(A.T @ A) @ A.T
-        expected_chisquare = np.trace((np.identity(x.shape[-1]) - P_phi) @ W @ cov @ W)
-        output.chisquare_by_expected_chisquare = chisquare / expected_chisquare
+        if kwargs.get('correlated_fit') is True:
```
fjosw, line 587: I would suggest to leave `output.chisquare_by_expected_chisquare` empty in this case

fjosw (changes requested):
Thanks for the features! I added a few comments. In particular the asserts in the tests will always pass as is_zero and is_zero_within_error have to be called as functions

s-kuberski:
Thanks for the improvements, I have implemented them and adapted the tests for the fit routines.

fjosw (approved):

PR #9, opened 2021-11-16
Bug-Fix: Do not apply filter if all deltas vanish, i.e., if Obs is zero

s-kuberski:
Removing a bug that was present for vanishing obs and irregular chains.

### DIFF
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -975,7 +975,10 @@ def _filter_zeroes(names, deltas, idl, eps=Obs.filter_eps):
             new_names.append(name)
             new_deltas[name] = np.array(nd)
             new_idl[name] = ni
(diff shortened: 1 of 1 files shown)


pyerrors/obs.py
```diff
@@ -975,7 +975,10 @@ def _filter_zeroes(names, deltas, idl, eps=Obs.filter_eps):
             new_names.append(name)
             new_deltas[name] = np.array(nd)
             new_idl[name] = ni
-    return (new_names, new_deltas, new_idl)
+    if new_names:
```
fjosw, line 978: Isn't the code identical for both cases?

s-kuberski: Indeed, I did not insert the correct piece in this version...

PR #10, opened 2021-11-17
Feature/irregular mc

s-kuberski:
This change drastically reduces the time needed for derived_observables in the case of irregular MC chains that have been merged. 

_filter_zeroes has been very slow. This has been improved by two changes:
- instead of np.isclose, we now use a simple > (cf. https://github.com/numpy/numpy/issues/16160)
- Obs.is_merged is now a dict. Therefore, we only have to filter in ensembles, where a merged happened.

Before, the filter time was creating a huge overhead, that was prohibitively expensive. Now, the time is still non-negligible, but the delay is acceptable. One could think about abandoning the feature all together and only do the filtering if the user explicitly asks for it. This would speed everything up but in rare cases, the error estimate might be wrong.


fjosw (commented):
Look very good! I added one small comment.

fjosw:
I just copied the code over to derived_array. I will have to get rid of this code duplication at some point.

PR #11, opened 2021-11-29
First implemenation of a json I/O

s-kuberski:
This feature contains a first implementation for the input from and output to a new .json.gz format. 

A workaround had to be applied to have indentation (and not to have all information in a single line) in combination with the [cnfg, deltas] entries printed in a single line.

This is up for discussion. The Zeuthen dobs format might be as efficient (an implementation of the corresponding I/O will follow). However, anyone with a json-implementation should be able to parse the new file format easily.

### DIFF
--- a/pyerrors/input/json.py
+++ b/pyerrors/input/json.py
@@ -7,10 +7,12 @@ import datetime
 from .. import version as pyerrorsversion
 import platform
 import numpy as np
+import warnings
 
 
-def dump_to_json(ol, fname, description='', indent=4):
-    """Export a list of Obs or structures containing Obs to a .json.gz file
+def create_json_string(ol, fname, description='', indent=1):
+    """Generate the string for the export of a list of Obs or structures containing Obs
+    to a .json(.gz) file
 
     Parameters
     -----------------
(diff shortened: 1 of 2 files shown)


fjosw:
The following code
```python
import numpy as np
import pyerrors as pe

tt1 = pe.Obs([np.random.rand(100)], ['t|r1'], idl=[range(2, 202, 2)])
tt2 = pe.Obs([np.random.rand(100)], ['t|r2'], idl=[range(2, 202, 2)])
tt3 = pe.Obs([np.random.rand(102)], ['qe'])

tt = tt1 + tt2 + tt3

tt.tag = 'Test Obs'

pe.input.json.dump_to_json([tt, tt], 'test', 'My testfile')

q = pe.input.json.load_json('test.json.gz')
```
throws the warning
```
VisibleDeprecationWarning: Creating an ndarray from ragged nested sequences (which is a list-or-tuple of lists-or-tuples-or ndarrays with different lengths or shapes) is deprecated. If you meant to do this, you must specify 'dtype=object' when creating the ndarray
  retd['deltas'] = np.array(retd['deltas'])
```
The results seems fine.

pyerrors/input/json.py
```diff
@@ -0,0 +1,332 @@
...
+    if prog and verbose:
+        print('Data has been written using %s.' % (prog))
+    if version and verbose:
+        print('Format version %s' % (version))
+    if np.any([who, date, host] and verbose):
+        print('Written by %s on %s on host %s' % (who, date, host))
+    description = d.get('description', '')
+    if description and verbose:
+        print()
+        print(description)
```
fjosw, line 306:
```suggestion
        print("Description: ", description)
```
Maybe make clear that this is the file description?

pyerrors/input/json.py
```diff
@@ -0,0 +1,332 @@
...
+            fname += '.gz'
+
+        fp = gzip.open(fname, 'wb')
+        fp.write(jsonstring.encode('utf-8'))
+    else:
+        fp = open(fname, 'w')
+        fp.write(jsonstring)
+    fp.close()
+
+    # this would be nicer, since it does not need a string but uses serialization (less memory!)
```
fjosw, line 199: Remove block comment

pyerrors/input/json.py
```diff
@@ -0,0 +1,332 @@
...
+
+    if gz:
+        if not fname.endswith('.gz'):
+            fname += '.gz'
+
+        fp = gzip.open(fname, 'wb')
+        fp.write(jsonstring.encode('utf-8'))
+    else:
+        fp = open(fname, 'w')
+        fp.write(jsonstring)
```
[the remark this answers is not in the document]
s-kuberski: I'm not sure. This could be done via 
fp = open(fname, 'w', encoding='utf-8')
but it's no fun to read the resulting *.json file using less because of all the additional printout. However, JSON encoders are required to be able to interpret utf-8 and therefore I/O would still work as expected. I do not intend to use plain json, anyways.

s-kuberski:
There has been a bug when parsing the deltas (and converting them to arrays of arrays to access the columns). The above problems should be fixed now. I included the problematic cases in the tests.

pyerrors/input/json.py
```diff
@@ -0,0 +1,328 @@
...
+        retd['idl'] = []
+        retd['deltas'] = []
+        retd['is_merged'] = {}
+        for ens in d:
+            for rep in ens['replica']:
+                retd['names'].append(rep['name'])
+                retd['idl'].append([di[0] for di in rep['deltas']])
+                retd['deltas'].append(np.array([di[1:] for di in rep['deltas']]))
+                retd['is_merged'][rep['name']] = rep.get('is_merged', False)
+        retd['deltas'] = np.array(retd['deltas'])
```
fjosw, line 230: The visible deprecation warning stems from this line as it can happen that the individual entries of the list do not have the same length. Is it necessary to cast to an `numpy.ndarray` here?

fjosw (commented):
Thanks for the bugfix, I still see the `VisibleDeprecationWarning` when I run my example.

pyerrors/input/json.py
s-kuberski: I don't see the warning, but this may be related to the python version. It is convenient to use the array, because I can access columns, but it should be possible to restructure the code, such that it can be removed.

fjosw: The warning appears from numpy 1.19 onward. And could stop the code from running in the future so I would suggest a workaround.

fjosw: I think just removing the array cast does not break the code. The iterations you are using to reconstruct the objects should also work with standard lists.

s-kuberski: Turns out that the newly inserted cast in line 228 is enough. I deleted line 230, as it is not necessary anymore. Does this resolve the issue?

fjosw: Yes that works!

PR #12, opened 2021-11-30
Feature/covobs

s-kuberski:
Introduces Covobs to pyerrors:

It is now possible to include auxiliary information, that is specified by a value and its variance [this could lead to confusion, since the error instead of the variance is chosen as argument for pseudo_Obs...] or by several values and a covariance matrix without the need to generate a pseudo-random distribution. Via automatic differentiation, the gradient of an Observable with respect to these auxiliary 'Covobs' is computed to machine precision and available at all times.

The feature still has to be implemented in derived_array and is therefore not compatible with anything in linalg.py. In all other functions, Obs including Covobs should be treated correctly.


fjosw:
I noticed a small bug which has to do with the fact that dictionaries like obs.e_dvalue are no longer sorted by key because of the covobs addition.
In the following code
```python
import numpy as np
import pyerrors as pe
covobs = pe.cov_Obs(1.13, 0.04, 'Zfactor')
my_obs = pe.Obs([np.random.normal(0.2, 0.002, 500)], ['test'])
full_obs = my_obs * covobs
full_obs.gamma_method()
full_obs.plot_piechart()
```
The error shares are wrongly attributed in Obs.plot_piechart as the routine assumes that the dictionaries are sorted. One simple solution could be to explicity loop over the dictionary keys instead of the list comprehension.

fjosw:
The routine Obs.is_zero() does not work as expected for covobs.
Example:
```python
import pyerrors as pe
covobs = pe.cov_Obs(0.0, 0.3, 'Zfactor')
covobs.is_zero()
> True
```

s-kuberski:
Thanks! I have fixed the issues.

PR #14, opened 2021-12-01
Resolves issue 

### LINKED ISSUE #13: Recover tags from arrays in json output
When I write an array of Obs with different tags in a json.gz file and extract it again all Obs have the same tag which is composed from the individual tags. How about using some delimiter in this case from which the original tags can be reconstructed?

Example:
```python
import numpy as np
import pyerrors as pe

tt1 = pe.Obs([np.random.rand(100)], ['t|r1'], idl=[range(2, 202, 2)])
tt1.tag = 'f1'
tt2 = pe.Obs([np.random.rand(100)], ['t|r1'], idl=[range(2, 202, 2)])
tt2.tag = 'f2'
tt3 = pe.Obs([np.random.rand(100)], ['t|r1'], idl=[range(2, 202, 2)])
tt3.tag = 'f3'

pe.input.json.dump_to_json(np.array([tt1, tt2, tt3]), 'test', 'My testfile')

q = pe.input.json.load_json('test.json.gz')

print([o.tag for o in q])

> ['f1\nf2\nf3\n', 'f1\nf2\nf3\n', 'f1\nf2\nf3\n']
```

s-kuberski:
Adjusted handling of tags: All are written, if any is None; Removed unnused parameter fname from create_json_string; Created a fail-save version for the replacement of quotation marks in workaround for delta lists.

A more elegant solution for the nice output, currently done via the deltalist class, should be possible by a modification of the JSON encoder.

Closes #13

### DIFF
--- a/pyerrors/input/json.py
+++ b/pyerrors/input/json.py
@@ -10,7 +10,7 @@ import numpy as np
 import warnings
 
 
-def create_json_string(ol, fname, description='', indent=1):
+def create_json_string(ol, description='', indent=1):
     """Generate the string for the export of a list of Obs or structures containing Obs
     to a .json(.gz) file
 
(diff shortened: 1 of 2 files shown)


fjosw:
Thanks!

PR #17, opened 2021-12-02
Feature/improved jsongz

fjosw:
Minor improvements to the new json.gz format
- Added Timezone information to date
- Added additional check for idl when writing list or array outputs
- Format improvements


s-kuberski (approved):
Looks good! Testing the idl for equality is definitely a good idea.

PR #24, opened 2021-12-07
Feature/covobs

s-kuberski:
This fixes possible problems that can appear, when observables are exported to an ASCII format and reimported again or when the gradient of covobs is set manually.

Changing cov and grad can now be done via routines, that do some basic checks. In principle, we could also hide those routines, since changing cov and grad is not needed anywhere and can lead to problems.

### DIFF
--- a/pyerrors/covobs.py
+++ b/pyerrors/covobs.py
@@ -20,19 +20,7 @@ class Covobs:
         grad : list or array
             Gradient of the Covobs wrt. the means belonging to cov.
         """
-        self.cov = np.array(cov)
-        if self.cov.ndim == 0:
-            self.N = 1
-            self.cov = np.diag([self.cov])
-        elif self.cov.ndim == 1:
-            self.N = len(self.cov)
-            self.cov = np.diag(self.cov)
-        elif self.cov.ndim == 2:
-            self.N = self.cov.shape[0]
-            if self.cov.shape[1] != self.N:
-                raise Exception('Covariance matrix has to be a square matrix!')
-        else:
-            raise Exception('Covariance matrix has to be a 2 dimensional square matrix!')
+        self.set_cov(cov)
         if '|' in name:
             raise Exception("Covobs name must not contain replica separator '|'.")
         self.name = name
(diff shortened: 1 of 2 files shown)


pyerrors/obs.py
```diff
@@ -1069,7 +1069,7 @@ def derived_observable(func, data, array_mode=False, **kwargs):
     for o in raveled_data:
         for name in o.cov_names:
             if name in allcov:
-                if not np.array_equal(allcov[name], o.covobs[name].cov):
+                if not np.allclose(allcov[name], o.covobs[name].cov, rtol=1e-14, atol=1e-14):
```
fjosw, line 1072: Is the tolerance maybe too small and could cause problems with rounding errors?

pyerrors/covobs.py
```diff
@@ -44,13 +33,43 @@ def __init__(self, mean, cov, name, pos=None, grad=None):
...
-            self.grad = np.array(grad)
+            self.set_grad(grad)
         self.value = mean
 
     def errsq(self):
         """ Return the variance (= square of the error) of the Covobs
         """
         return float(np.dot(np.transpose(self.grad), np.dot(self.cov, self.grad)))
+
+    def set_cov(self, cov):
```
fjosw, line 47: I would vote for hidding the `set` routines

fjosw (commented):
Looks very good! I have two minor comments.

pyerrors/obs.py
s-kuberski: Good question. I got problems because exact equality was needed before. Maybe rtol=1e-14 is enough to capture differences that are based on finite machine precision? Or even less?
I was thinking about small numbers, but a small atol could lead to problems for entries that are significantly larger than one.

pyerrors/covobs.py
s-kuberski: Agreed. One could even think about using a double underscore for the variables, to make (accidental) changes even more difficult. What do you think?

pyerrors/obs.py
fjosw: I usually use the standard parameters for `allclose` (`rtol=1e-05`, `atol=1e-08`) based on the assumption that this agreement is sufficient for our purposes. When the tolerance approaches machine precision I feel like there could be special cases which cause problems. But this is just a feeling, I never encountered such a case.

s-kuberski: You are right, any explicit change of the covariance matrix would already be captured using the less strict bounds. Probably it's fine.

pyerrors/covobs.py
fjosw: Double underscores could (possibly) cause problems within classes (see [PEP8](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables)). The main reason why I advocate for leading underscores in functions which are irrelevant for users is that the auto generated documentation also omits them. For the instance attributes like `_value` my idea was to inform the user when attempting to overwrite `value`.

PR #26, opened 2021-12-09
Feature/import json string

fjosw:
I propose to split the function `load_json` up into two parts, one that reads the file and one that deciphers the json string and reconstructs the objects.


s-kuberski:
Looks good!

PR #30, opened 2021-12-14
covobs in json I/O

s-kuberski:
Included covobs in the json file format using the additional entry "cdata" that is written right after "data". It is also possible to export/import Obs that are pure covobs, i.e., that do not have any deltas.


fjosw:
Thanks!

PR #31, opened 2021-12-20
Feature/json corr

fjosw:
I added the Correlator class to the json input and output.

### DIFF
--- a/pyerrors/input/json.py
+++ b/pyerrors/input/json.py
@@ -7,6 +7,7 @@ import datetime
 import platform
 import warnings
 from ..obs import Obs
+from ..correlators import Corr
 from ..covobs import Covobs
 from .. import version as pyerrorsversion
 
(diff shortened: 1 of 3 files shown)


pyerrors/input/json.py
```diff
@@ -173,6 +174,19 @@ def write_Array_to_dict(oa):
...
 
+    def write_corr_to_dict(my_corr):
+        if my_corr.N > 1:
+            raise Exception("Output for N>1 not implemented")
+        front_padding = next(i for i, j in enumerate(my_corr.content) if j)
+        back_padding_start = front_padding + next((i for i, j in enumerate(my_corr.content[front_padding:]) if not j), my_corr.T)
+        my_dict = write_List_to_dict([o[0] for o in my_corr[front_padding:back_padding_start]])
+        my_dict['type'] = 'Corr'
+        my_dict['padding_front'] = front_padding
+        my_dict['padding_back'] = my_corr.T - back_padding_start
```
s-kuberski, line 185: I don't think that it is a good idea to add new keywords for every data type, that is implemented to the json I/O. The additional information about the padding could also be put in / read from tag or description (the latter is no generic solution, though). How about using ["tag|1", "tag|2", ..., "tag|N-1"], or something similar to indicate the time slice position, for the list of tags instead? This would be perfectly valid within the format.
Since the padding should not be a dominant part of the correlator, you could also just print out the None entries, right?

In any case, it would be nice if one could parse json files without dealing with too many special cases. If Corr had the same entries as List, one could simply parse it as list, if one did not use pyerrors.

s-kuberski (commented):
I have added some comments regarding the general question, if we want to stick to the previously defined json entries or if we want to add new entries for each structure.

pyerrors/input/json.py
fjosw: My idea was to construct the 1D correlator output as a superset of a list. Anyone not familiar with this extension can still read the file in the list format and would just loose out on the correlator tag and the padding. I am open to other solution, though.

The general problem that I see with the new format is that it is rather complex with 331 lines of code and is based on many special cases.

s-kuberski: The same would be true, if you write out None for time slices that are not considered and indicate the padding in the tag, or am I missing something? The main idea behind my suggestions was to keep compatibility with the "List" format in not introducing new key words but putting everything in the structure, that is already present.

I guess that you either have more lines of code in the I/O or more lines in the definition of the format. I still have to write the format definition down, but if we want to change something beforehand, we should do it now. 

Tbh I don't really now how to cut the format without loosing some of the information that we want to put in. Concerning the special cases: I tried to design the format such that everything that is in there is derived from a list. After reading the list, everything else is just parsing the list structure to a pre-defined structure. If you want to import and export many different structures, I don't know how to get around some parsing.

fjosw:
Closed in favor of #34

PR #32, opened 2021-12-21
Feature/input 2.0

Xenror:
Replaced the sfcf_read methods with just one method for all of them.
Replaced the read_qtop methodwith one that directly reads from openQCD outputs.
Implemented idl handling for the afore metioned methods.


pyerrors/input/openQCD.py
```diff
@@ -343,3 +345,209 @@ def _read_array_openQCD2(fp):
...
+        specify the exact files that need to be read from path, pratical if e.g. only one replicum is needed
+    names: list
+        Alternative labeling for replicas/ensembles. Has to have the appropriate length
+    """
+    #dtr_cnfg = 4# was ist das denn hier?
+    #one could read L from the header in case of sfQCD
+    #c = 0.35
+    known_versions = ["1.0","1.2","1.4","1.6","2.0", "sfqcd"]
+    version = "1.2"
+    if "version" in kwargs:
```
fjosw, line 389: I would propose to make `version` a parameter in line 350. This would make lines 388-390 redundant.

pyerrors/input/openQCD.py
```diff
@@ -343,3 +345,209 @@ def _read_array_openQCD2(fp):
...
+        for q in Q_round:
+            if full:
+                tmp.append(q)
+            else:
+                if int(q) == target:
+                    tmp.append(1.0)
+                else:
+                    tmp.append(0.0)
+        
+        truncated_file = file[:-7] #as seen in previous examples, this could lead to some weird behaviour... maybe -7 fixes this.
```
fjosw, line 525: Why does `[:-7]` help here?

pyerrors/input/openQCD.py
```diff
@@ -343,3 +345,209 @@ def _read_array_openQCD2(fp):
...
+        #keyword "names" prevails over "ens_name"
+        if not "names" in kwargs:
+            rep_names.append(truncated_file[:idx] + '|' + truncated_file[idx:])
+        else:
+            names = kwargs.get("names")
+            rep_names = names
+        deltas.append(np.array(tmp))
+
+        
+    result = Obs(deltas, rep_names)
```
fjosw, line 552: Ideally, the `Obs` object would also obtain the `idl` of the data.

pyerrors/input/__init__.py
```diff
@@ -4,3 +4,4 @@
 from . import misc
 from . import openQCD
 from . import sfcf
+from . import utils
```
fjosw, line 7: I think importing utils in the individual modules should be sufficient, the corresponding functions do not need to be available in the global namespace.

pyerrors/input/sfcf.py
```diff
@@ -6,17 +6,41 @@
...
+    b2b: bool
+        if True, read a time-dependent boundary-to-boundary correlation function
+    single: bool
+        if True, read time independent boundary to boundary correlation function
+    names: list
+        Alternative labeling for replicas/ensembles. Has to have the appropriate length
+    ens_name : str
+        replaces the name of the ensemble
+    version: str
+        version of SFCF, with which the measurement was done. if the compact output option (-c) was spectified, append a c to the version (e.g. "1.0c")
```
fjosw, line 37: Ideally, one could also specify `a` for the append mode.

pyerrors/input/sfcf.py
```diff
@@ -29,266 +53,210 @@ def read_sfcf(path, prefix, name, **kwargs):
...
-
+    if "replica" in kwargs:
+        reps = kwargs.get("replica")
+    if "files" in kwargs:
+        files = kwargs.get("files")
+
+    #due to higher usage in current projects, compact file format is default
+    compact = True
+    #get version string
+    version = "1.0"
```
fjosw, line 69: Also here I would make `version` a parameter with standard value.

fjosw (changes requested):
Thanks for all your work! I left a few minor comments. The linter came back with several errors but these should be easy to fix. I would also propose to remove all redundant comments, print statements etc. To integrate with the current test workflow the file `input_test.ipynb` would have to be replaced by a python file with the appropriate `pytest` syntax. The existing tests should be a good guide here.

In general I would propose to simplify the functions and remove as many parameters as possible as we do not have to keep any backward compatibility.

pyerrors/input/sfcf.py
```diff
@@ -29,266 +53,210 @@ def read_sfcf(path, prefix, name, **kwargs):
...
-        first_cfg = int(re.findall(r'\d+', sub_ls[0])[-1])
-
-        last_cfg = len(sub_ls) + first_cfg - 1
-
-        for cfg in range(1, len(sub_ls)):
-            if int(re.findall(r'\d+', sub_ls[cfg])[-1]) != first_cfg + cfg:
-                last_cfg = cfg + first_cfg - 1
+        if "files" in kwargs:
+            sub_ls = kwargs.get("files")
+            sub_ls.sort(key=lambda x: int(re.findall(r'\d+', x)[-1]))
```
s-kuberski, line 125: Does this work in combination with the automatic name recognition? You could filter first, before you try to detect the name from the content of ls, to throw out all of the stuff, that you don't like anyways.

pyerrors/input/sfcf.py
```diff
@@ -29,266 +53,210 @@ def read_sfcf(path, prefix, name, **kwargs):
...
+                            floats = list(map(float, line.split()))
+                            if version == "0.0":
+                                deltas[k-start][i][cnfg] = floats[im]
+                            else:
+                                deltas[k - start][i][cnfg] = floats[1 + im - single]
+
+    if "check_configs" in kwargs:
+        print("Chekcing for missing configs...")
+        che = kwargs.get("check_configs")
+        if not (len(che) == len(idl)):
```
s-kuberski, line 252: You could also do something like 
```python
if len(set(che).symmetric_difference(idl)) > 0
```
to catch cases, where the two lists are not equal but have the same length.

s-kuberski (commented):
Nice work. When you have finished the changes to read_qtop, I can implement the changes that are necessary to read the files from the 1.6 version of sfqcd.

pyerrors/input/openQCD.py
jkuhl-uni: Compared to before (I think it was something like `file.split(".")[0]`) this version is more stable when it comes to filenames containing a ".", e.g. T32L32_b3.80_k0.13... , which appeared in my analyses more than often.

pyerrors/input/__init__.py
jkuhl-uni: Okay, fair point

pyerrors/input/sfcf.py
jkuhl-uni: Yes, this was my idea as well. When starting to implement that feature, I will go with this parameter

pyerrors/input/openQCD.py
```diff
@@ -343,3 +345,209 @@ def _read_array_openQCD2(fp):
...
+    steps: int
+        step size of measurements
+    L: int
+        spatial length of the lattice in L/a. HAS to be set if version != sfqcd, since openQCD does not provide this in the header
+    r_start: list
+        offset of the first ensemble, making it easier to match later on with other Obs
+    r_stop: list
+        last ensemble that needs to be read
+    r_meas_start: list
+        offset of the first measured ensemble, if there is any
```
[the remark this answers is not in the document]
jkuhl-uni: Yes, sorry

pyerrors/input/openQCD.py
```diff
@@ -343,3 +345,209 @@ def _read_array_openQCD2(fp):
...
+    prefix:
+        prefix of the measurement files, e.g. <prefix>_id0_r0.ms.dat
+    c:
+        ???
+    dtr_cnfg:
+        ???
+    target: int
+        specifies the topological sector to be reweighted to (default 0)
+    full: bool
+        if true read the charge instead of the reweighting factor.
```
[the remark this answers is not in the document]
jkuhl-uni: This is actually a great idea. The second method will be quite easy, and the first one is de-cluttered a little.

pyerrors/input/sfcf.py
jkuhl-uni: I think I do not understand this comment. These two are lists of lists in which the configurations per replicum are listed, this part just checks if there is a list to check for each replicum. `set(che).symmetric_difference(idl)` is almost what is done here, together with `check _idl()` for each rep separately.

pyerrors/input/openQCD.py
jkuhl-uni: On the other hand, of course `[:-7]` is not stable towards changes in the file extension.

pyerrors/input/sfcf.py
jkuhl-uni: If I understand this correctly, the answer is: The name recognition is dependent on the replicum names, "files" just filters the content in each replicum, so yes, this should work as far as I am aware (but I tested if `files` without `names` nonetheless).
In any case, a feature that is not implemented yet is something like files that can be different for each replicum, this was rather low on my priority list for now...

pyerrors/input/openQCD.py
jkuhl-uni: Okay, version is now a parameter with default value `"1.2"`

pyerrors/input/sfcf.py
jkuhl-uni: `version` is now also a parameter here, with default value `1.0`.

pyerrors/input/openQCD.py
```diff
@@ -343,3 +345,209 @@ def _read_array_openQCD2(fp):
...
+    Parameters
+    ----------
+    path:
+        path of the measurement files
+    prefix:
+        prefix of the measurement files, e.g. <prefix>_id0_r0.ms.dat
+    c:
+        ???
+    dtr_cnfg:
+        ???
```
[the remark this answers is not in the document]
jkuhl-uni: Thank you for the clarification, this also makes some other things easier (e.g. this is redundant with paramter steps, which was not clear to me)

pyerrors/input/openQCD.py
jkuhl-uni: Done, hopefully in the right way.

pyerrors/input/openQCD.py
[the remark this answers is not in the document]
jkuhl-uni: This is in fact not redundant as I thought, now it seems to me as if there is a difference between the oQCD1.6 and oQCD2.0 output that I cannot find in the output methods...

pyerrors/input/openQCD.py
```diff
@@ -343,3 +345,209 @@ def _read_array_openQCD2(fp):
...
+    """Read qtop format from given folder structure.
+
+    Parameters
+    ----------
+    path:
+        path of the measurement files
+    prefix:
+        prefix of the measurement files, e.g. <prefix>_id0_r0.ms.dat
+    c:
+        ???
```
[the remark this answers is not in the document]
jkuhl-uni: Okay, thank you for the clarification.

pyerrors/input/openQCD.py
```diff
@@ -343,3 +346,241 @@ def _read_array_openQCD2(fp):
...
+
+        truncated_file = file[:-7]
+        print(truncated_file)
+        idl_start = 1
+        
+        if "r_start" in kwargs:
+            Q_round = Q_round[r_start[rep]:]
+            idl_start = r_start[rep]
+        if "r_stop" in kwargs:
+            Q_round = Q_round[:r_stop[rep]]
```
fjosw, line 519: I think the slicing here does not work properly if both `r_start` and `r_stop` are defined. I would propose something like this:
```python
if "r_start" in kwargs:
    first_cnfg = r_start[rep]
    idl_start = r_start[rep]
else:
    first_cnfg = None
    idl_start = 1
if "r_stop" in kwargs:
    last_cnfg = r_stop[rep]
else:
    last_cnfg = None

Q_round = Q_round[first_cnfg: last_cnfg]
idl_stop = idl_start+len(Q_round)
```

pyerrors/input/sfcf.py
```diff
@@ -29,266 +53,306 @@ def read_sfcf(path, prefix, name, **kwargs):
...
+                else:
+                    new_names.append(entry[:idx] + '|' + entry[idx:])
+        else:
+            
+            for exc in ls:
+                if not fnmatch.fnmatch(exc, prefix + '*.'+name):
+                    ls = list(set(ls) - set([exc]))
+            ls.sort(key=lambda x: int(re.findall(r'\d+', x)[-1]))
+            for entry in ls:
+                myentry = entry.removesuffix("."+name)
```
fjosw, line 140: This is a feature which was only added in python 3.9. As we support all version >=3.6 this would have to be changed. See for example https://stackoverflow.com/a/1038845

pyerrors/input/sfcf.py
```diff
@@ -6,17 +6,41 @@
 import re
 import numpy as np  # Thinly-wrapped numpy
 from ..obs import Obs
+from . import utils
 
-
-def read_sfcf(path, prefix, name, **kwargs):
-    """Read sfcf C format from given folder structure.
+def read_sfcf(path, prefix, name, quarks='.*', noffset=0, wf=0, wf2=0, version = "1.0c", **kwargs):
+    """Read sfcf c format from given folder structure.
```
fjosw, line 12: Documentation of the parameters `path`, `prefix` and `name` seems to be missing.

fjosw:
Thanks for all your work, I will merge the pull request and we can test it a bit on the develop branch before putting together the release candidate for version 2.0.

PR #33, opened 2021-12-23
Some more changes to corr

JanNeuendorf:
- GEVP method can return lists of eigenvector by solving the GEVP at multiple timeslices. The ordering is done according to arXiv:2004.10472 [hep-lat] 
- The projection method can deal with those lists 
- Constructor for Hankel matrix from a single corr 
- typechecks allow for complex-content 
- .real and .imag work on corrs
- But everything else with CObs does not yet work.

### DIFF
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -3,7 +3,7 @@ import numpy as np
 import autograd.numpy as anp
 import matplotlib.pyplot as plt
 import scipy.linalg
-from .obs import Obs, reweight, correlate
+from .obs import Obs, reweight, correlate, CObs
 from .misc import dump_object
 from .fits import least_squares
 from .linalg import eigh, inv, cholesky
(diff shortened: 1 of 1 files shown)


fjosw:
Hi Jan, thanks for your work! The tests fail because I changed the Corr.dump method today and added `kwargs` such that the user can specify a path. This should be easily fixed my merging the develop branch. The errors of the linter should be rather straight forward.
Thanks also for including the complex type into the Corr objects. For my purposes it is not too important to have all the features also available for complex correlators, so we could merge the pull request in its current state once all tests are passed.

fjosw:
After our discussion today I added a few additional points:
- [ ] For some of the new methods the docstrings are still missing
- [ ] Possible tests GEVP
- [ ] It would be good to have an example for the GEVP and multidimensional correlators

fjosw:
I just noticed that the print feature does not work for multi dimensional correlators. Maybe this is something you could also work on.

PR #34, opened 2022-01-18
Feature/json corr nd

fjosw:
Here is a new version of the json import export for the correlator class based on three dimensional arrays. The implementation should work for one and multidimensional arrays. The additional metadata (padding and correlator tag) is saved in an additional entry to the list of tags. As separator for this metadata I used `|`.


s-kuberski (approved):
Looks fine, thanks for adapting the code.

PR #39, opened 2022-01-27
An update to the documentation for the correlator class.

JanNeuendorf:
I put in a little example for the matrix-correlators.


fjosw:
Thanks a lot!

PR #47, opened 2022-01-28
This commit contains four small changes.

JanNeuendorf:
1. Some docstrings for the GEVP stuff
2. The example data was updated to be loaded with the new Corr.dump() format (Nones instead of padding)
3. You can now init a Corr from an array of other (non-matrix) corrs. This might be useful for reading in data.
You can read in one smearing (or similar) at a time, save it as a corr and then decide to build a matrix from that later. It checks for most of the things, that can go wrong (i hope) and then converts the input to the accepted list format.

4. The GEVP  t_s parameter is now a keyword argument, because it is not needed in all cases.


JanNeuendorf:
I do not know, why it does not allow the merge. I think changing the indentation on your code-block confused it.

fjosw:
I think the merge conflict stems from the binary files. I modified them already yesterday to agree with the new format so there should be no need to update them.

fjosw:
I just had a look at the conflict, it actually has to do with the patch for m_eff in combination with padding. I think you should merge the develop branch first to resolve the conflict.

fjosw:
I think you are missing commit 90717ba85d6d9715dd4aae016e17da6995e08906 in your history which conflicts with your changes.

Issue #46, opened 2022-01-27
corr.dump method does not handle prange


JanNeuendorf:
It looks like the prange parameter does not get preserved after saving and reloading. 
This would be very important to me, but i do not want to change it myself out of fear of breaking the format.

s-kuberski:
I would really like the format to survive in its current form..., JSON allows to write dictionaries inside the tag and/or description entries. We could also rethink the way, the padding is written to the tag and instead do something like
```python
tag = {'tag': some_tag, 'padding': [pl, pu], 'prange': [ra, rb]}
```
In this way, it would be easy to recover any kind of information, that you like to export together with the correlator.

JanNeuendorf:
I do like this idea, because it would allow me to easily preserve other parameters i set. (Like t_0 for the GEVP for example) . The only concern i have is that now some keys are reserved for this internal functions. 
For example i would like to keep it so that corr.plateau() defaults to the prange unless specified otherwise. 
If we do this, we might enforce 1. the tag being a dictionary and 2. inform the user about keys with special meanings.

s-kuberski:
I don't understand the problem with 'special keys'. I did not intend to have the possibility to mess around with the way the tag is written. Everything, that would be included the above format would be based on a property of the correlator class. 
If you like to store a lot of additional information together with your data, you can already do so by using the tag and/or the description. It is perfectly valid to use dictionarys instead of strings here.
However, I would still think of the change from above to make sure that all properties of the correlator class can be written and recovered. Your additional information would be within ```some_tag``` in my snippet.

JanNeuendorf:
Okay, so it is not me, who sets the dictionary, but just the dump() method that makes it a dictionary before saving it? 


I also want to mention, how much i love the json format for this. Today i build a little API that just gives back the json for a requested object. It is really easy to fetch an obs or correlator from a remote machine. 

1. If Fabian allows the dependency on "requests", i would allow the load() functions to directly accept urls. 
2. We could have some example data on GitHub pages so you do not have to download it.

s-kuberski:
Yes! My idea of translating the tag and the additional information to a dictionary for the export of the correlator class was intended to be hardcoded within the dump routine, such that all the properties of the class can be coherently exported and imported.
You would have the freedom to write whatever you want inside the correlator tag.

JanNeuendorf:
Okay, that sounds great. 
Will you do it or should i try?

fjosw:
[...]

PR #48, opened 2022-01-28
Changes in the correlator init, GEVP and docstrings & tag format includes dict and prange

JanNeuendorf:
Okay, i repulled and redid the changes in the correlators file only. 
Are you sure, that i should not update my example data?

### DIFF
--- a/pyerrors/input/json.py
+++ b/pyerrors/input/json.py
@@ -403,13 +403,15 @@ def import_json_string(json_string, verbose=True, full_output=False):
         if isinstance(o.get('tag'), list):  # supports the old way
             taglist = o.get('tag')  # This had to be modified to get the taglist from the dictionary
             temp_prange = None
-        else:
+        elif isinstance(o.get('tag'), dict):
             tagdic = o.get('tag')
             taglist = tagdic['tag']
             if 'prange' in tagdic:
                 temp_prange = tagdic['prange']
             else:
                 temp_prange = None
+        else:
+            raise Exception ("The tag is not a list or dict")
 
         corr_tag = taglist[-1]
         tmp_o = o
--- a/pyerrors/input/json.py
+++ b/pyerrors/input/json.py
@@ -186,6 +186,11 @@ def create_json_string(ol, description='', indent=1):
             dat['tag'].append(corr_meta_data)
         else:
             dat['tag'] = [corr_meta_data]
+        taglist = dat['tag']
+        dat['tag'] = {}  # tag is now a dictionary, that contains the previous taglist in the key "tag"
+        dat['tag']['tag'] = taglist
+        if my_corr.prange is not None:
+            dat['tag']['prange'] = my_corr.prange
         return dat
 
     if not isinstance(ol, list):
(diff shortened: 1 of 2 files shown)


fjosw:
The only thing I did is adjusting the tags to the new format, i.e. removing the excessive `0|0|`. Did you make any other changes? Or in other words. Does the example still work properly?

fjosw:
Now the tests are failing because you simply overwrote my patch 90717ba85d6d9715dd4aae016e17da6995e08906 please include these changes in your version for the tests to pass.

JanNeuendorf:
Yes, i just tested it, the examples still run on my machine. If this ever breaks because of an update to the format, i can just recreate them

fjosw:
Thanks! Just a small suggestion: Do you see any problem in merging the two parameters `return_list` and `sorting`? It appears to me as if just having one parameter could give the same functionality.

JanNeuendorf:
Okay, so it would be 
return_list=none to return a single value 
and return_list="Eigenvalue" for the sorting? 
sounds a bit confusing, but it would work. 
How about you allow the merge and i will change it for the next commit?

JanNeuendorf:
Okay, i named the argument sorted_list 
it is None by default, so it does not return a list and you can set it to a string to give the sorting method.

JanNeuendorf:
Okay, i changed the get_Corr_from_dic() and write_Corr_to_dict() to replace the old list "tag" with a dictionary. The dictionary has a key that is also named "tag", which holds the list, which is replaced. 
This sounds confusing, but @s-kuberski  explained it nicely. The dump method will always use the dictionary and include the prange in it, if available. The load method checks, if the tag is a list or a dic and proceeds accordingly. We can decide later, if we want to break this compatibility for simplicity. 
I think, this is actually a good solution, because we could include any optional parameters Corr could get in the future in this dictionary.

fjosw:
Looks good, maybe @s-kuberski can also have a look before I merge the PR.
Closes #46

pyerrors/input/json.py
```diff
@@ -395,7 +400,17 @@ def get_Array_from_dict(o):
         return np.reshape(ret, layout)
 
     def get_Corr_from_dict(o):
-        taglist = o.get('tag')
+        if isinstance(o.get('tag'), list):  # supports the old way
+            taglist = o.get('tag')  # This had to be modified to get the taglist from the dictionary
+            temp_prange = None
+        else:
```
fjosw, line 406: For safety reasons I would suggest to check whether `o.get('tag')` is a `dict` and raise an Exception if the tag is neither `list` nor `dict`.

fjosw:
I would also advocate to add an additional test to pyerrors/tests/io_test.py in which it is checked whether the new dictionary tag and specifically the `prange` is correctly reconstructed from a `json` file.

pyerrors/input/json.py
JanNeuendorf: Yes, i added the exception

s-kuberski: I agree. Too bad, that we already have to care about backwards compatibility, but I hope, we don't have to include too many of such cases. The tag-dictionary will certainly help in this regard.

JanNeuendorf:
I agree that tests are needed. But maybe we should all get on one version and **then** decide how to test it

pyerrors/input/json.py
JanNeuendorf: I guess as long as it is pre 2.0, Fabian could decide to break the format. 
One could easily update data into the new format.

s-kuberski:
Looks good to me. In principle, we are now in the position to remove the leading and trailing None's from the JSON file using a 'padding' key in the dict, but this is not too important.

fjosw:
For the future I would suggest that all new features have to come with basic tests. In this case the additional functionality can be easily tested by adding a line or two, I can do that on the fly.

PR #49, opened 2022-01-31
feat!: Implemented improved first and second derivatives for the corresponding methods of the Corr class.

fjosw:
The parameter which distinguishes between the variant is now a string instead of a bool which may break current analyses.

### DIFF
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -460,25 +460,16 @@ class Corr:
 
         return (self + T_partner) / 2
 
-    def deriv(self, symmetric=True):
+    def deriv(self, variant="symmetric"):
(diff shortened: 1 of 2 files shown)


s-kuberski (commented):
This is a useful feature! Does it make sense to implement the backward derivative as well, to complete the set?
The tests could include checks of the validity, e.g., by checking that the second derivative is equal to two times the symmetric derivative (this is the case, right?) or that backward + forward = symmetric.

fjosw:
Good idea, I added the backward derivative and additional tests. The relation which I test for the second derivative is
```python
corr.deriv('forward').deriv('backward') == corr.second_deriv()
```

s-kuberski (approved):
Thanks, that looks good!

PR #54, opened 2022-02-03
Several improvements for read_rwms and extract_t0

s-kuberski:
Both routines now support:
- idl for the resulting Obs
- Stepsize >1 in units of configurations (this is not strictly necessary for RWFs, since it is allowed to have a superset of the configs of Obs in reweight)

- It is now possible to visualize the fit for t0.
- extract_t0 now also supports the keywords "files" and "names" that allows to use a subset of the replica and ensembles that are not named according to a classical ens + r + rep convention.


s-kuberski:
Actually, the introduction of idl can be considered as bugfix, since it was possible to provide r_start != 1 and this was not reflected in the idl of the Obs.

fjosw:
Very good, thanks!

PR #55, opened 2022-02-04
Ensure correctness of idl in RWFs and t0

s-kuberski:
We now read the configuration numbers in the case of RWFs to ensure the correctness of idl.

In the case of t0, this is more difficult, because we only know the number of the HMC trajectory or SMD cycle. It is now assumed, that the first measurement belongs to the first config and that a possible offset is due to thermalization. A warning is thrown, if an offset is present. There is a new optional keyword to change this behavior and to assume that no thermalization is present in the MC run.

We do not correctly handle cases where the distance between two measurements is not equal to the distance between two configurations. It is hard to catch all cases and I would not know how to handle the idl consistently, if the distance between to measurements is smaller than the config distance. For now, I added a comment in the specification of the function, such that users can handle this issue themselves.

### DIFF
--- a/pyerrors/input/openQCD.py
+++ b/pyerrors/input/openQCD.py
@@ -226,7 +226,12 @@ def extract_t0(path, prefix, dtr_read, xmin,
     The data around the zero crossing of t^2<E> - 0.3
     is fitted with a linear function
     from which the exact root is extracted.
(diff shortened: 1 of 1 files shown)


fjosw:
Thanks for taking care of this. I agree with you that there are far too many special cases so warnings are a good way forward.

PR #57, opened 2022-02-08
Rewritten large parts of Qtop routines, making it compatible with sfqcd

s-kuberski:
This was a lot more work than expected. The openQCD and sfqcd versions of the Wilson/Gradient flow measurements are only vaguely similar. Maybe, it would have been better to write a separate routine for each of them. On top of this, there are many different possible combinations of trajectory numbers in the file and configuration numbers.

However, it should work now with both versions. The Zeuthen flow works with sfqcd, I don't know if this is also part of the openQCD-2.0 measurements. In this case, it has to be added.

I also cleaned up the method(s) to generate a projection factor from the charge. 

@fjosw Please go though the keywords and check if you think, that anything could be removed. There are many of them, but I think all of them might be necessary.
@jkuhl-uni Please verify, that this code gives the same results as the old implementation in your case. On my side, everything seemed to be fine.

### DIFF
--- a/pyerrors/input/openQCD.py
+++ b/pyerrors/input/openQCD.py
@@ -217,8 +217,7 @@ def read_rwms(path, prefix, version='2.0', names=None, **kwargs):
     return result
 
 
-def extract_t0(path, prefix, dtr_read, xmin,
-               spatial_extent, fit_range=5, **kwargs):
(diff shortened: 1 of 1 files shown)


fjosw:
Thanks a lot! I had a look at the keyword arguments and I don't think there is anything that we can easily get rid off. I will merge the code now, in case there is any incompatibility just open a new PR @jkuhl-uni

PR #59, opened 2022-02-09
Bugfix in openQCD routines: r_stop was excluded instead of being the …

s-kuberski:
…last element

### DIFF
--- a/pyerrors/input/openQCD.py
+++ b/pyerrors/input/openQCD.py
@@ -196,6 +196,8 @@ def read_rwms(path, prefix, version='2.0', names=None, **kwargs):
             else:
                 try:
                     r_stop_index.append(configlist[-1].index(r_stop[rep]) + 1)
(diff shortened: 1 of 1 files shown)


s-kuberski:
Wait, that might have been too fast.

s-kuberski:
It is really hard to track all the special cases. We definitely need some test routines for these input routines. I did not appropriately handle the case, where r_stop was equal to the last config. This is now also taken care of and I hope, that everything is fine.

fjosw:
Thanks for spotting the bug. I agree that tests would be really reassuring. I am a little afraid of bloating up the repo with many test files but I guess we can find a solution to this.

s-kuberski:
I'll create some small tests based on a short 4^4 chain. The RWFs have a negligible size. The gradient flow measurements however would need about 200K (for .gfms.dat - Zeuthen flow and .ms.dat - Wilson flow for t0).

PR #60, opened 2022-02-09
Yet another small change to openQCD. Added explicit tests for reweighting

s-kuberski:
I had to change the idl creation another time. Now I am more certain, that everything works as it should. I created tests using
- sfqcd-1.6, one RWF with one source
- openQCD-2.0 two RWFs: TM with 2 Hasenbusch factors and 2 sources each, RHMC with 1 source

Here, I discovered that the trajectory numbers instead of the configuration numbers are printed in the case of the sfqcd on-line measurements. I applied a fix to reconstruct the idl correctly.

Adding tests for the gradient flow measurements would add about 0.3M of data for minimal working examples. I don't know if this is worth the effort. For the RWFs, I was more worried.


fjosw:
Very good! I think 0.3M is reasonable so adding those test would be a good idea if you can spare the time.

PR #64, opened 2022-02-10
fix: erroneous replica separator in json format caught and corrected

fjosw:
With this little fix I am covering a very special case which should never happen with files written by pyerrors. This fix explicitly makes sure that the replica separator `|` is enforced in case another program wrote a json file with a different separator. This fix would imply that the separator `|` is not part of the format specification and would allow other implementations to use other separators.

### DIFF
--- a/pyerrors/input/json.py
+++ b/pyerrors/input/json.py
@@ -307,10 +307,15 @@ def import_json_string(json_string, verbose=True, full_output=False):
             retd['is_merged'] = {}
             for ens in d:
                 for rep in ens['replica']:
-                    retd['names'].append(rep['name'])
+                    rep_name = rep['name']
+                    if len(rep_name) > len(ens["id"]):
+                        tmp_list = list(rep_name)
+                        tmp_list[len(ens["id"])] = "|"
+                        rep_name = ''.join(tmp_list)
+                    retd['names'].append(rep_name)
                     retd['idl'].append([di[0] for di in rep['deltas']])
                     retd['deltas'].append(np.array([di[1:] for di in rep['deltas']]))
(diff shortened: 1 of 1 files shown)


s-kuberski:
This is a good idea! However, it would lead to problems, if no separator is written, at all. In the special case of:
ens_name = 'H101'
rep_names = ['H101a', 'H101b']
the creation of the Obs, and therefore loading the JSON file, would fail. 
You could circumvent this by doing something like
```python
                        tmp_list = tmp_list[:len(ens["id"])] + ["|"] + tmp_list[len(ens["id"]):]
```
This would be fail-safe. In the case that a different separator has been written, it would now be part of the definition of the rep_names. The user could take care of this afterwards. (I was always thinking of having a method that safely renames replica inside an Obs)

s-kuberski:
Of course, my snipped does not work at all for the case, where the separator "|" has been written! You would need a check of
```python
tmp_list[len(ens["id"])] == "|"
```

fjosw:
Thanks for spotting this, I think my last commit should account for this special case.

s-kuberski (changes requested):
The newly introduced if always evaluates to False, because you introduce the separator before. This should be
```python
tmp_list = list(rep_name)
if rep_name[len(ens["id"])] != "|":
    tmp_list = list(rep_name)
    tmp_list = tmp_list[:len(ens["id"])] + ["|"] + tmp_list[len(ens["id"]):]
    rep_name = ''.join(tmp_list)
else:
    tmp_list[len(ens["id"])] = "|"
    rep_name = ''.join(tmp_list)
```

fjosw:
The code worked for test cases I constructed. Isn't the `else` case you constructed redundant as it does not change `rep_name`?

PR #72, opened 2022-02-15
bug fix: read methods are in fact closing files now

jkuhl-uni:
Bug fixed that some files aren't closed in read method. Forgot to close the files that are used to look where the fitting correlator can be found in the files.


fjosw:
Thanks for the quick fix!

PR #73, opened 2022-02-16
Feature/json io

s-kuberski:
I had another look at our JSON routines. The changes reduce the RAM requirements considerably in when writing to JSON and a bit when loading a JSON file.

In the former case this is done via using an iterative JSON encoder and by removing an unnecessary intermediate dict from RAM.

In the latter case, I dropped the creation of an intermediate string, by using json.loads instead of json.load. This means that as much memory is saved, as the unzipped JSON file requires.

### DIFF
--- a/pyerrors/input/json.py
+++ b/pyerrors/input/json.py
@@ -6,6 +6,7 @@ import datetime
 import platform
 import warnings
 import re
+import gc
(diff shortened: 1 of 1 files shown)


fjosw:
Great, thanks!

PR #74, opened 2022-02-16
Added JSON schema for validation and documentation purposes

s-kuberski:
This JSON schema may be used to check whether a file is valid with respect to the JSON file format that we have defined. This validation is not able to find any type of bug, e.g., if the layout and the number of entries per config inside a structure match. However, anyone can use it to check whether his or her own implementation of the file format is written in the correct way.

As side effect, this is a self-contained documentation of the file format. I intend to write a more concise definition for the pyerrors documentation. If we like to have this schema as self-contained explanation, some more information on the different types of entries will have to be given in the "description" tags. (Probably we can use the same text for both tasks).

Feel free to comment on this and to suggest improvements.

### DIFF
--- a/examples/json_schema.json
+++ b/examples/json_schema.json
@@ -1,0 +1,343 @@
+{
+ "$schema": "http://json-schema.org/draft-07/schema",
+ "$id": "https://github.com/fjosw/pyerrors/tree/develop/examples/json_schema.json",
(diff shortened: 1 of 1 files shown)


fjosw:
Good idea! I think a plain text explanation of the format in the documentation would still be very useful.

PR #75, opened 2022-02-17
repair

JanNeuendorf:
The problem with the matrix-init for correlators was fixed and there is a new method called

<code>remove_outliers()</code> in the obs module that helps with removing faulty configurations.
<code>corr.plateau()</code> now also has <code>auto_gamma</code> , but it defaults to <code>False</code>.


s-kuberski:
I have a few thoughts on ```remove_outliers```:
- Is it a good idea to have this kind of method in Obs? For a new user, this could imply that it is perfectly normal to exclude some configurations from the MC chain. However, this is certainly not the case. In my opinion, you would need a solid understanding why a configuration might be problematic (exceptionally low eigenmodes, exceptional configurations in quenched calculations). Even if you have it, you would have to be sure that these configs are really "bad" in the sense that they are catastrophically large outliers and not part of the distribution, you try to sample. It is dangerous to mess around with the path integral and it should not be easy to do this.
- If you really want such method in this package, I would propose a different ansatz: 
1.) Put the method in misc.py such that it takes some time to find it.
2.) Rewrite the method such that it finds exceptional configurations, but don't remove them from the Obs.
(3.) Don't remove the config, at all. The easiest way to get rid of such configs is to reweight them to 0. This point is debatable, but it is easy to reweight, while it is more involved to throw out configs that enter an Obs.).
4.) Take care: One special Obs might be ill-defined one one config, while another one is totally fine. Would you like to mix these in your analysis or do you exclude the config from you chain for all observables?
- I think that it is a good thing to have a method that inspects the properties of the distribution (apart from what is already in ```plot_history```. If you have a non-Gaussian distribution and you use the median instead of the mean, you should not use the standard deviation as definition of the width of your distribution. Instead you could proceed as described in footnote 1 of [hep-lat/0512021], which would give you something like this:
```python
def get_width(o):
    ddata = [list(o.deltas[name] + o.r_values[name]) for name in o.names]
    data = []
    for d in ddata:
        data += d
    N = len(data)
    data = np.sort(data)
    nbase = int(.683 * N)
    wl = []
    for i in range(N-nbase):
        wl.append(data[i + nbase] - data[i])
    width = .5 * np.min(wl)
    print('Width: %1.3f vs. %1.3f' % (width, np.std(data)))
    return width
```
of course with the correct percentage for your Nsigma.
- You should rename your keyword to ```threshold```.

fjosw:
I agree with Simon, removing configurations from a sampled distribution is a very dangerous game and we should not encourage users to do so.

fjosw:
Concerning the `get_width` functions: I think the width should be determined for every ensemble separately as the underlying distributions may vary.
It is probably a good idea to talk about non-Gaussian distributions in one of our future meetings.

JanNeuendorf:
1. I agree that it is a dangerous method. But really every method is dangerous, if not used correctly. Maybe the name should be changed. When i say faulty configurations, i do not mean statistical outliers. I mean values, which are many orders of magnitudes off due to some input-output error. That is, why i set the default value to 5 sigma. (it could be more) 
2. I put it in obs.py, since it takes and returns an Obs. It could be in misc. or utils 
3. I do not think it is necessary to have separate methods to find and remove those values. I wrote the method, so it could be used as a check as well as a cleaning-tool. 
4. I already checked, that the sigma calculated looks like the standard deviation of the delta-distribution 
5. With regards to reweighting vs. removing. It comes back to 1. This is only for values, which are wrong due to some bug. So i want to remove them completely. 
6. Yes, spelling is hard.

fjosw:
In my opinion any I/O errors should be taken care of on the level of the I/O routines. I have to say that I see more harm than use in adding such a function. As Simon said this might give users the impression that removing outliers is a valid part of an analysis workflow.

Please send individual pull requests for bug fixes and new features in the future so we can discuss them separately.

s-kuberski:
> Concerning the `get_width` function: I think the width should be determined for every ensemble separately as the underlying distributions may vary. It is probably a good idea to talk about non-Gaussian distributions in one of our future meetings.

Your are right. I had this method lying around, but it was intended for single-ensemble Obs only. I agree that it could be beneficial to think about non-Gaussian distributions.

s-kuberski:
> 1. I agree that it is a dangerous method. But really every method is dangerous, if not used correctly. Maybe the name should be changed. When i say faulty configurations, i do not mean statistical outliers. I mean values, which are many orders of magnitudes off due to some input-output error. That is, why i set the default value to 5 sigma. (it could be more)
> 
In this case, a routine that checks for these outliers (as it is part of your routine), would be very valuable. Some kind of Obs-sanity-check could be helpful, in general. But I would argue that the next step would be to repair your I/O and not to throw out these configs. This should *not* become part of the main branch.
>     4. I already checked, that the sigma calculated looks like the standard deviation of the delta-distribution
> 
If you have some tailed distributions, I would expect it to fail, anyways.

JanNeuendorf:
I think you are missing the point of finding the sigma. I only need some approximation here.  
As for I/O problems being handled in the I/O methods. Yes, the only time, one would use this, is after reading in data. 
But why would i write some structure to hold and analyze the data first, when Obs already exists.

JanNeuendorf:
i give up

fjosw:
I wrote a test which expects an Exception when a correlator is initialized with observables defined on different configurations. This test seems to fail with your most recent additions
```
    def test_corr_exceptions():
        obs_a = pe.Obs([np.random.normal(0.1, 0.1, 100)], ['test'])
        obs_b = pe.Obs([np.random.normal(0.1, 0.1, 99)], ['test'])
        with pytest.raises(Exception):
>           pe.Corr([obs_a, obs_b])
E           Failed: DID NOT RAISE <class 'Exception'>
```

*Once more: Please use separate branches and pull requests for individual features and bug fixes so we can discuss and merge them separately.*

PR #77, opened 2022-02-21
Changes to Eigenvalue

JanNeuendorf:
`GEVP` method now uses `scipy.eigh()` and applies matrix_symmetric().
`Eigenvalue` takes the same arguments as `GEVP` and projects automatically.

### DIFF
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -266,10 +266,11 @@ class Corr:
             if (self.content[t0] is None) or (self.content[ts] is None):
                 raise Exception("Corr not defined at t0/ts")
             G0, Gt = np.empty([self.N, self.N], dtype="double"), np.empty([self.N, self.N], dtype="double")
+            symmetric_corr = self.matrix_symmetric()
             for i in range(self.N):
                 for j in range(self.N):
-                    G0[i, j] = self.matrix_symmetric().content[t0][i, j].value
-                    Gt[i, j] = self.matrix_symmetric().content[ts][i, j].value
+                    G0[i, j] = symmetric_corr.content[t0][i, j].value
+                    Gt[i, j] = symmetric_corr[ts][i, j].value
 
             sp_vecs = _GEVP_solver(Gt, G0)
             sp_vec = sp_vecs[state]
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -302,7 +302,7 @@ class Corr:
         return all_vecs
 
     def Eigenvalue(self, t0, ts=None, state=0, sorted_list=None):
-        vec = self.GEVP(self, t0=t0, ts=ts, state=state, sorted_list=sorted_list)
+        vec = self.GEVP(t0, ts=ts, state=state, sorted_list=sorted_list)
         return self.projected(vec)
 
     def Hankel(self, N, periodic=False):
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -7,7 +7,6 @@ import scipy.linalg
 from .obs import Obs, reweight, correlate, CObs
 from .misc import dump_object, _assert_equal_properties
 from .fits import least_squares
-# from .linalg import eigh, inv, cholesky
 from .roots import find_root
 
 
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -7,7 +7,7 @@ import scipy.linalg
 from .obs import Obs, reweight, correlate, CObs
 from .misc import dump_object, _assert_equal_properties
 from .fits import least_squares
-from .linalg import eigh, inv, cholesky
+# from .linalg import eigh, inv, cholesky
 from .roots import find_root
 
 
@@ -268,8 +268,8 @@ class Corr:
             G0, Gt = np.empty([self.N, self.N], dtype="double"), np.empty([self.N, self.N], dtype="double")
             for i in range(self.N):
                 for j in range(self.N):
-                    G0[i, j] = self.content[t0][i, j].value
-                    Gt[i, j] = self.content[ts][i, j].value
+                    G0[i, j] = self.matrix_symmetric().content[t0][i, j].value
+                    Gt[i, j] = self.matrix_symmetric().content[ts][i, j].value
 
             sp_vecs = _GEVP_solver(Gt, G0)
             sp_vec = sp_vecs[state]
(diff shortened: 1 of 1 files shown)


pyerrors/correlators.py
```diff
@@ -268,8 +268,8 @@ def GEVP(self, t0, ts=None, state=0, sorted_list=None):
             G0, Gt = np.empty([self.N, self.N], dtype="double"), np.empty([self.N, self.N], dtype="double")
             for i in range(self.N):
                 for j in range(self.N):
-                    G0[i, j] = self.content[t0][i, j].value
-                    Gt[i, j] = self.content[ts][i, j].value
+                    G0[i, j] = self.matrix_symmetric().content[t0][i, j].value
+                    Gt[i, j] = self.matrix_symmetric().content[ts][i, j].value
```
fjosw, line 272: It would be more efficient to call `matrix_symmetric` only once before the loops instead of symmetrizing every single time.

pyerrors/correlators.py
```diff
@@ -301,24 +301,9 @@ def GEVP(self, t0, ts=None, state=0, sorted_list=None):
...
-            if self.content[t] is None:
-                newcontent.append(None)
-            else:
-                Gt = G.content[t]
-                M = Li @ Gt @ LTi
-                eigenvalues = eigh(M)[0]
-                eigenvalue = eigenvalues[-state]
-                newcontent.append(eigenvalue)
-        return Corr(newcontent)
+    def Eigenvalue(self, t0, ts=None, state=0, sorted_list=None):
```
fjosw, line 304: Please add a brief docstring for this method.

pyerrors/correlators.py
```diff
@@ -1081,7 +1066,7 @@ def _sort_vectors(vec_set, ts):
 
 
 def _GEVP_solver(Gt, G0):  # Just so normalization an sorting does not need to be repeated. Here we could later put in some checks
-    sp_val, sp_vecs = scipy.linalg.eig(Gt, G0)
+    sp_val, sp_vecs = scipy.linalg.eigh(Gt, G0)
```
fjosw, line 1069: By now I am not sure anymore whether `eigh` actually checks if the matrix is positive semi definite. Do we need an additional safeguard here or not?

pyerrors/correlators.py
```diff
@@ -7,7 +7,7 @@
 from .obs import Obs, reweight, correlate, CObs
 from .misc import dump_object, _assert_equal_properties
 from .fits import least_squares
-from .linalg import eigh, inv, cholesky
+# from .linalg import eigh, inv, cholesky
```
fjosw, line 10: Please remove this line.

fjosw (changes requested):
Thanks for the quick changes, I have a few minor comments

fjosw:
I just saw that `test_matrix_corr` fails which might suggests that something goes wrong with the new GEVP.
What I test is:
```python
corr_aa = _gen_corr(1)
corr_ab = _gen_corr(0.5)

corr_mat = pe.Corr(np.array([[corr_aa, corr_ab], [corr_ab, corr_aa]]))

vec_0 = corr_mat.GEVP(0, 0)
corr_0 = corr_mat.projected(vec_0)

assert np.all([o == 0 for o in corr_0 - corr_aa])
```
After the changes `corr_0` and `corr_aa` seem to be different while they agreed for the old version. Is this a sensible test?

JanNeuendorf:
I will take a look at the test later.

pyerrors/correlators.py
JanNeuendorf: Yes, it does. "If eigenvalue computation does not converge, an error occurred, or b matrix is not definite positive. Note that if input matrices are not symmetric or Hermitian, no error will be reported but results will be wrong."

fjosw:
I think this has to do with the ordering of the states. With the new version `corr_1` is equal to `corr_aa`. I am not sure if the test I am imposing is suitable here. We could ditch it for a better test.

JanNeuendorf:
What you are testing here is the problem 
**Mv=λMv** so **λ=1** and every **v** is a solution. I think the only reason, that the test worked in the first place is, that it starts "guessing" with (0,1) and (1,0) (and `eigh` does not). 
Even if it gives the desired **v**s, it can not sort by **λ**. Does this makes sense or am i missing something?

fjosw:
Yes, I think you are right. Can you construct a more meaningful test for the GEVP method?

pyerrors/correlators.py
```diff
@@ -1081,7 +1066,7 @@ def _sort_vectors(vec_set, ts):
 
 
 def _GEVP_solver(Gt, G0):  # Just so normalization an sorting does not need to be repeated. Here we could later put in some checks
-    sp_val, sp_vecs = scipy.linalg.eig(Gt, G0)
+    sp_val, sp_vecs = scipy.linalg.eigh(Gt, G0)
     sp_vecs = [sp_vecs[:, np.argsort(sp_val)[-i]] for i in range(1, sp_vecs.shape[0] + 1)]
```
fjosw, line 1070: I think the ordering is implicitly done in the `eigh` call and is redundant. Can you confirm that?

s-kuberski: This is a little more involved. First of all, `eigh` sorts in descending order, but in lattice applications, you want to have the lowest states. So you have to invert this. On top of this, a matrix of eigenvectors is returned and you need to get the columns of this matrix, in order to have the eigenvectors. 
So you go from 
```
array([[ 4.08809560e+01,  4.71498284e+00,  6.77498937e-01,
        -1.66951604e+01],
       [-4.58757541e+02, -5.50717440e+02, -3.16503614e+01,
         1.99051675e+03],
       [ 5.38668672e+02,  5.26371852e+02,  7.59161722e+01,
        -5.84411783e+03],
       [-2.53763739e+01,  4.70594189e+02,  2.34206426e+02,
         4.96899729e+03]])
```
to 
```
[array([  -16.69516037,  1990.51675171, -5844.11782914,  4968.99728957]),
 array([  0.67749894, -31.65036143,  75.91617222, 234.20642623]),
 array([   4.71498284, -550.71744041,  526.37185249,  470.59418928]),
 array([  40.88095603, -458.75754148,  538.66867207,  -25.37637386])]
```
(based on the data in the example. Is this correct, @JanNeuendorf?

s-kuberski: In general, this functions needs a doc-string (although it is an internal function). The order of the returned eigenvectors should be specified. Also, we need to specify that the results are incorrect, if the matrices are non-symmetric or explicitly catch those cases.

fjosw:
Can we find a (temporary) solution for this issue today?

PR #79, opened 2022-02-22
Adds the documentation of the Covobs class

s-kuberski:


fjosw:
Thanks! That should be the last missing piece in the documentation.

PR #81, opened 2022-02-23
_merge_idx now returns sorted lists

s-kuberski:
This fixes a bug when combining Obs that are defined on irregular MC chains:

The following snipped 
```python
import pyerrors as pe
a = pe.pseudo_Obs(1, .1, 'a', samples=10)
a.idl['a'] = range(4, 15)
b = pe.pseudo_Obs(1, .1, 'a', samples=151)
b.idl['a'] = range(4, 608, 4)
ol = [a, b]
o = (ol[0] - ol[1]) / (ol[1])
```
throws an exception
```
...pyerrors/obs.py in _expand_deltas_for_merge(deltas, idx, shape, new_idx)
    996         if idx == new_idx:
    997             return deltas
--> 998     ret = np.zeros(new_idx[-1] - new_idx[0] + 1)
    999     for i in range(shape):
   1000         ret[idx[i] - new_idx[0]] = deltas[i]

ValueError: negative dimensions are not allowed
```

The reason is that _expand_deltas_for_merge expects new_idx to be sorted. This had not been ensured before.

### DIFF
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -971,7 +971,7 @@ def _merge_idx(idl):
             idstep = min([idx.step for idx in idl])
             return range(idstart, idstop, idstep)
 
-    return list(set().union(*idl))
+    return sorted(set().union(*idl))
 
 
 def _expand_deltas_for_merge(deltas, idx, shape, new_idx):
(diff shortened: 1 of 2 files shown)


fjosw:
Thanks for spotting that! Can you include that minimal example as a test which expects an `Exception`?

s-kuberski:
That depends. It should not happen in a normal workflow anymore, because the list should be properly sorted. I'll add the above snippet as test (that should work) and adapt the function specifications.

PR #83, opened 2022-03-01
Reworked covariance

fjosw:
I removed the `fits.covariance_matrix` function and adjusted `covariance` to operate on a list of observables and return a matrix.
I also reworked the way the covariance is estimated and I believe that the new estimator ensures positive semi-definiteness of the covariance matrix up to machine precision. The `covariance` function now also allows for a direct visualization of the corresponding (normalized) correlation matrix. I also added additional tests for the covariance.

### DIFF
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -1348,7 +1348,7 @@ def covariance(obs, visualize=False, correlation=False, **kwargs):
 
     Notes
     -----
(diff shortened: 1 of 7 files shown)


pyerrors/obs.py
```diff
@@ -1332,21 +1332,60 @@ def correlate(obs_a, obs_b):
...
+        plt.draw()
+
+    if correlation is True:
+        return corr
+    else:
+        return cov
+
+
+def _covariance_element(obs1, obs2):
+    """Estimates the uncorrelated covariance of two Obs objects."""
```
s-kuberski, line 1387: Uncorrelated covariance sounds a bit strange. Maybe "covariance of two Obs objects, neglecting autocorrelation" or something similar?

s-kuberski (commented):
Looks good.

pyerrors/obs.py
fjosw: Agreed

s-kuberski (approved):

PR #84, opened 2022-03-03
Simplification and speed up of the computation of e_N in _covariance_element

fjosw:
I think the line `np.sum(np.ones_like(idl_d[r_name]))` is equivalent to `len(idl_d[r_name])` but the second option is simpler and faster. Can you confirm that @s-kuberski ?


s-kuberski:
This should be correct. Your exchange is equivalent for ```list```, ```numpy.array```, ```range```.

PR #85, opened 2022-03-03
Replaced inverse by the pseudo inverse in the calculation of chi_exp

fjosw:
In some tests I ran today I noticed that the matrix `A.T @ A` which has to be inverted for our estimation of the expected chisquare can be numerically very close to singular in some cases. One (at least temporary) solution would be to replace the inverse by the Moore-Penrose pseudo inverse which would filter out the directions corresponding to the almost vanishing singular values. In my test the estimator for the expected chisquare became negative when using the standard inverse but was quite close to the "true" correlated chisquare value when using the pseudo inverse. As I am not using this feature on a regular basis I wanted to ask for other opinions.

### DIFF
--- a/pyerrors/fits.py
+++ b/pyerrors/fits.py
@@ -249,7 +249,7 @@ def total_least_squares(x, y, func, silent=False, **kwargs):
         new_jac = np.concatenate((fused_row1, fused_row2), axis=1)
 
         A = W @ new_jac
-        P_phi = A @ np.linalg.inv(A.T @ A) @ A.T
+        P_phi = A @ np.linalg.pinv(A.T @ A) @ A.T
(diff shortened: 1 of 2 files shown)


s-kuberski:
I have seen negative expected chisquares in the past when scanning fit ranges and just ignored those fits. It is good to know that you might have found the origin. I did not expect to see problems in the inversion here. 
I think your solution is a good one since it stabilizes the procedure in corner cases and should not change anything when ```A.T @ A``` can be safely inverted. 

In the long run, it would be good to find out what triggers those problems.

fjosw:
I think the problem occurs when the matrix `A` does not have full rank. The projection matrix H=A @ (A.T @ A)^(-1) @ A.T should be a projector and thus idempotent (trace(H) = rank(A)). In these edge cases this property is not fulfilled and I think truncating vanishing singular values restores the projection property.

PR #86, opened 2022-03-04
Prevent division by zero by setting values of gamma_div which are smaller than 1 to 1

fjosw:
When working with spaced and irregular data it can occur that `gamma` and `gamma_div` are both numerically zero for some windowsizes. In the ratio of 0 / 0 odd rounding errors can occur. With this fix I propose to set all elements of `gamma_div` that are smaller than 1 to 1 which should not affect the standard error estimation.


s-kuberski:
I had not thought about these cases. Your solution looks good!

PR #87, opened 2022-03-07
feat: implemented eigenvalue smoothing method of hep-lat/9412087

fjosw:
I implemented the eigenvalue smoothing method described in hep-lat/9412087. I am not sure how useful this method is but having it available may be worthwhile.

I am also unsure whether this should still enter version 2.0.


s-kuberski:
I think it does not hurt to have something like this at hand for the fits. In the long run, one could think about providing the correlation matrix (or its inverse) directly to the fit routine. In this way, one can easily modify it or create, e.g., block diagonal correlation matrices.
If we are going to rewrite the fit routines for version 2.1, this feature could also enter at that point. In this case, we would not have to think about how to access the routine within the fit routine.

fjosw:
I agree, we should give the API for the fits a bit more thought.

> In this case, we would not have to think about how to access the routine within the fit routine.

In the current implementation the keyword arguments are propagated so one could call a fit function with `correlated_fit=True` and `smooth=8` to do a correlated fit with a smoothed correlation matrix. This is not explained in the documentation, yet.

PR #89, opened 2022-03-21
Precision of is_zero

fjosw:
I propose to increase the standard values for the precision which is required for two observables to be considered equal. The formula which is used within numpy is `absolute(a - b) <= (atol + rtol * absolute(b))` (https://numpy.org/doc/stable/reference/generated/numpy.isclose.html).

### DIFF
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -443,7 +443,7 @@ class Obs:
         """
         return self.is_zero() or np.abs(self.value) <= sigma * self._dvalue
 
-    def is_zero(self, rtol=1.e-5, atol=1.e-8):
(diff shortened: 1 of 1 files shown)


s-kuberski:
I agree, the value for ```atol``` might have been a bit high to compare with 0. The value for rtol should be completely irrelevant, if you compare with 0, right? Because you have 
```
absolute(a - b) <= (atol + rtol * absolute(b))
absolute(0 - b) <= (atol + rtol * absolute(b))
absolute(b) <= atol + rtol * absolute(b)
(1 - rtol) * absolute(b) <= atol
```
in all the cases that matte within the routine.

fjosw:
I agree, `rtol` should be completely irrelevant here.

PR #90, opened 2022-04-07
Fix/covariance idl

fjosw:
I fixed a bug in `covariance` which occurred for some combinations of idls (see the new `test_covariance_idl`). I think the fix is just to replace `expand_deltas` within `covariance` by `obs._expand_deltas_for_merge`. The difference between the two is only that the latter collapses the expanded data back into the provided idl. Can you take a closer look at this @s-kuberski ?

### DIFF
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -996,7 +996,7 @@ def _intersection_idx(idl):
             idstep = max([idx.step for idx in idl])
             return range(idstart, idstop, idstep)
 
-    return sorted(set().intersection(*idl))
+    return sorted(set.intersection(*[set(o) for o in tt]))
 
 
 def _expand_deltas_for_merge(deltas, idx, shape, new_idx):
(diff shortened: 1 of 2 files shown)


fjosw:
This PR fixes the explicit error but I am not sure if `covariance` produces the correct behavior for two `Obs` defined on different idls. Shouldn't we base the calculation of the covariance only on the intersect of configurations instead of the union?

fjosw:
I implemented the functions `_intersection_idx` and `_collapse_deltas_for_merge` and modified `covariance` to conform with the aforementioned property that the correlation of two observables should not change if data is added on disjunct configurations (the covariance still does). I hope that the behavior for observables defined on the same set of samples is unchanged (all tests were passed). Please take a careful look.

fjosw:
I am afraid this solution allows for negative eigenvalues of the covariance in cases when observables are defined on different idls. We should discuss this issue in detail.

PR #95, opened 2022-04-26
added in code to impliment the cutoff method

yup790:
The method is explained in arxiv:1101.2248v8  section 4.2

### DIFF
--- a/pyerrors/fits.py
+++ b/pyerrors/fits.py
@@ -462,6 +462,17 @@ def _standard_fit(x, y, func, silent=False, **kwargs):
 
     if kwargs.get('correlated_fit') is True:
         corr = covariance(y, correlation=True, **kwargs)
+        
+        
+        if kwargs['cutoff']:
+            cutoff=kwargs.get('cutoff')
+            eigvals=np.linalg.eig(corr)[0]
+            eigvecs=np.array(np.linalg.eig(corr)[1:])[0]
+            max_eigval=np.max(np.abs(eigvals))
+            eigvals[np.abs(eigvals)< max_eigval*cutoff]=0
+            corr=np.linalg.inv(eigvecs.T)@np.diag(eigvals)@eigvecs.T    
+
+
         covdiag = np.diag(1 / np.asarray(dy_f))
         condn = np.linalg.cond(corr)
         if condn > 0.1 / np.finfo(float).eps:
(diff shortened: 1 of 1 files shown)


pyerrors/fits.py
```diff
@@ -462,6 +462,17 @@ def _standard_fit(x, y, func, silent=False, **kwargs):
 
     if kwargs.get('correlated_fit') is True:
         corr = covariance(y, correlation=True, **kwargs)
+        
+        
+        if kwargs['cutoff']:
+            cutoff=kwargs.get('cutoff')
+            eigvals=np.linalg.eig(corr)[0]
+            eigvecs=np.array(np.linalg.eig(corr)[1:])[0]
```
s-kuberski, line 470: You call ```np.linalg.eig``` twice. It would be better to do this once before assigning to ```eigvals``` and ```eigvecs```

pyerrors/fits.py
```diff
@@ -462,6 +462,17 @@ def _standard_fit(x, y, func, silent=False, **kwargs):
 
     if kwargs.get('correlated_fit') is True:
         corr = covariance(y, correlation=True, **kwargs)
+        
+        
+        if kwargs['cutoff']:
```
s-kuberski, line 467: This functionality is not explained in the docstring of ```least_squares```. Please add an explanation such that users are aware of the feature.

pyerrors/fits.py
```diff
@@ -462,6 +462,17 @@ def _standard_fit(x, y, func, silent=False, **kwargs):
...
+            cutoff=kwargs.get('cutoff')
+            eigvals=np.linalg.eig(corr)[0]
+            eigvecs=np.array(np.linalg.eig(corr)[1:])[0]
+            max_eigval=np.max(np.abs(eigvals))
+            eigvals[np.abs(eigvals)< max_eigval*cutoff]=0
```
s-kuberski, line 472: I am not totally sure about the method, but as I understood it, you would like to set the inverse of the eigenvalues to zero if the eigenvalue is too large, right? If you set the eigenvalue to zero, the inversion below has to fail, I guess.

s-kuberski (changes requested):

s-kuberski:
Hi. Thanks for contributing! Please look at the test section in https://github.com/fjosw/pyerrors/blob/develop/CONTRIBUTING.md to get rid of the errors in the checks. 
I added some comments to your code. In general, I think that this is an interesting feature. As I understand it, the meaning of the fit quality (as given by the reduced chisquare) is not really clear anymore, when the covariance matrix is modified. Therefore, it would be interesting to see this in combination with the expected chisquare.
In the long run (for pyerrors 2.1) the fit routines are going to be restructured. At that point it will make sense to be able to provide a (modified) covariance matrix to the fit such that it can be used in every routine.

PR #96, opened 2022-04-28
Implemented the pruning of large correlation matrices by the solution…

s-kuberski:
… of a GEVP at early times

This may help to allow for a solution of the GEVP at late time slices based on a large operator basis. At the same time, the correctness of the GEVP is not in question.

Unfortunately, the method GEVP does not allow to return all eigenvectors at the same time such that an unnecessary number of inversions has to be performed. This could be improved by changing ```GEVP``` (the docstring of that method is not totally instructive) and adapting ```prune``` afterwards.

### DIFF
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -1103,6 +1103,67 @@ class Corr:
 
         return self._apply_func_to_corr(return_imag)
 
+    def prune(self, Ntrunc, tproj=3, t0proj=2, basematrix=None):
+        r''' Project large correlation matrix to lowest states
+
+        This method can be used to reduce the size of an (N x N) correlation matrix
+        to (Ntrunc x Ntrunc) by solving a GEVP at very early times where the noise
+        is still small.
+
+        Parameters
+        ----------
+        Ntrunc: int
+            Rank of the target matrix.
+        tproj: int
+            Time where the eigenvectors are evaluated, corresponds to ts in the GEVP method.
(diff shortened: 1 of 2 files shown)


pyerrors/correlators.py
```diff
@@ -1103,6 +1103,67 @@ def return_imag(obs_OR_cobs):
...
+        correlation matrix and to remove some noise that is added by irrelevant operators.
+        This may allow to use the GEVP on $G(t)$ at late times such that the theoretically motivated
+        bound $t_0 \leq t/2$ holds, since the condition number of $G(t)$ is decreased, compared to $C(t)$.
+        '''
+
+        if self.N == 1:
+            raise Exception('Method cannot be applied to one-dimensional correlators.')
+        if basematrix is None:
+            basematrix = self
+        elif basematrix.N == 1:
```
fjosw, line 1145: I think this line is redundant to `1149` as a more general check.

fjosw:
Thanks, this looks very good! I left a small comment in the code. Is there a reference that could be mentioned in the docstring?

s-kuberski:
There is not really a reference except for our paper that needs to be written in the near future (we have not explained this in 2111.04693)... People in the literature use variants of this method but I think they assume that the diagonal elements of the pruned matrix are already dominated by the respective states. By using the GEVP on the pruned matrix, you don't have to rely on this assumption. I don't think that anyone has done this, so far.

PR #97, opened 2022-05-02
Feature/dobs

s-kuberski:
This pull request adds the possibility to read and write using the Zeuthen pobs (for Obs based on single ensembles) and dobs formats. This is not meant to replace the JSON format but as interface to the ```dobs``` analysis package.


s-kuberski:
I hope that I have spotted all small things that one has to take care about, but it is easiest to find them when this code is used actively.

s-kuberski:
Alright, this adds the need for the ```lxml``` package, I forgot about this. Maybe we have to discuss it further.

PR #98, opened 2022-05-09
Use rapidjson instead of json for I/O

s-kuberski:
This change from json to rapidjson helps to achieve an order of magnitude of speedup when writing large JSON files to disk. At the same time, the memory footprint is smaller. The option ```write_mode=json.WM_SINGLE_LINE_ARRAY``` allows to get rid of the ugly workaround for having a human readable format, if indent!=0. 

This would add the dependence on ```python-rapidjson``` (I have not added this to the ```setup.py```, because I like to discuss it first). We could insert something like 
```python
try:
    import rapidjson as json
except ImportError:
    import json
```
to allow the user to do the export using the ```json``` module and to improve performance by installing ```rapidjson```. This would need some small adjustments in ```create_json_string``` to handle both cases appropriately.

### DIFF
--- a/pyerrors/input/json.py
+++ b/pyerrors/input/json.py
@@ -1,4 +1,4 @@
-import json
+import rapidjson as json
 import gzip
 import getpass
 import socket
@@ -6,7 +6,6 @@ import datetime
 import platform
 import warnings
 import re
-import gc
 import numpy as np
 from ..obs import Obs
 from ..covobs import Covobs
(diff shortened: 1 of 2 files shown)


s-kuberski:
> Sounds good, I don't see a problem with adding `rapidjson` as a dependency, it looks like that library is being actively maintained. Does this also affect reading JSON files?

The performance for reading JSON files does not really depend on the JSON parser because most of the time is spent building the ```Obs```.

s-kuberski:
> I had a quick look at the `rapidjson` documentation and it looks like https://github.com/ijl/orjson is even more performant. Did you also have a look at that?

I had a look at this, as well. The main difference to `rapidjson` is that the output is a binary, such that this would have to be decoded, if you want to save the python string anywhere. Of course, when you export to `json.gz`, you want a binary. So one would have to do some unproblematic reshuffling in `dump_to_json`. At the same time, `rapidjson` has the nice output that removes the need for my self-written routines. With `orjson` this would not really be possible without a degradation of the performance (or having either everything in one line or a line per entry in the deltas). I did not see any significant difference in speed when switching between `rapidjson` and `orjson`, at least for my test dictionary with 2000 Obs. Since the performance seemed to be similar and the nice output comes out of the box with `rapidjson`, I went for the latter. It is basically a drop-in replacement for `json`, whereas `orjson` requires more changes.

It might be possible thought that there are cases where orjson is a bit faster. By now, when exporting to a 3.1 MB large .json.gz file, 60% is spent in 'fp.write' and only 40% in `create_json_string`. Within `create_json_string`, only one third of the time is spent in `json.dumps`. Therefore, the performance of the export does not really improve anymore, when the parses gets faster. This might depend a bit on the properties of the `Obs` that one is exporting and the file system, but the general picture should be true. Compared to the previous implementation, this is a significant shift of the workload.

I could also implement `orjson` (I had done it anyways for my tests), but I think `rapidjson` should perform equally well and I like looking at files with `less` ;)

PR #99, opened 2022-05-12
GEVP rehaul

fjosw:
I propose a few changes to the `Corr.GEVP` method: With my patch the GEVP is not evaluated for t <= t0, instead `None` is appended in case of the `Eigenvalue` and `Eigenvector` sorting methods. The `Eigenvector` sorting method can now cope with None entries.


s-kuberski:
Looks fine for me. It is a good idea to set those values to None, since they should not be used to do physics.

PR #100, opened 2022-05-16
GEVP rehaul 2

fjosw:
I propose to make two additional **breaking** changes to `Corr.GEVP`. With my patch
- the `GEVP` method now returns all vectors instead of just the ones for a specified state. Since we compute all vectors anyways this saves unnecessary overhead and also eliminates the argument `state`. If only one state is of interest the corresponding vectors can be obtained via indexing
```
C.GEVP(t0=2)[0]  # Ground state vectors
C.GEVP(t0=2)[:3]  # Vectors for the lowest three states
```
- the argument `sorted_list` was renamed to `sort` which feels like a more meaningful name.

I also simplified the `_solve_GEVP` helper function, the functionality should be unchanged. I added a unit test which verifies this.

Opinions?

Depends on PR #99

### DIFF
--- a/examples/06_gevp.ipynb
+++ b/examples/06_gevp.ipynb
@@ -168,9 +168,9 @@
     }
    ],
    "source": [
-    "vec = matrix_V1V1.GEVP(t0=3, ts=6 ,state=0, sorted_list=None)\n",
-    "assert len(vec) == matrix_V1V1.N\n",
-    "print(vec)"
+    "vecs = matrix_V1V1.GEVP(t0=3, ts=6, sort=None)\n",
+    "assert len(vecs[0]) == matrix_V1V1.N\n",
+    "print(vecs[0])"
    ]
   },
   {
@@ -202,7 +202,7 @@
     }
    ],
    "source": [
-    "matrix_V1V1.projected(vec).m_eff().show(comp=single_smearing.m_eff(), auto_gamma=True)"
+    "matrix_V1V1.projected(vecs[0]).m_eff().show(comp=single_smearing.m_eff(), auto_gamma=True)"
    ]
   },
   {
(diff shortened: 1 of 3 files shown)


s-kuberski:
In my opinion, it should definitely be possible to get all eigenvectors with one call. One could also do this by returning all vectors if state=None, thereby keeping compatibility with older releases (this would not have been my preferred solution for a new routine). 

I don't have an opinion concerning ```sorted_list```. You could look for the keyword ```sorted_list```, save the value in ```sort``` and throw a deprecation warning, to keep backwards compatibility for some time. 

In general, I think that one should specify the sorting pattern (descending) in the docstrings and a we should have a docstring for `_GEVP_solver` that includes the restrictions for the matrix to guarantee correctness of the method.

fjosw:
I followed up on Simon's suggestions and reintroduced `sorted_list` and `state` as deprecated `kwargs`. I also extended the docstrings of `GEVP` and `_GEVP_solver`.

JanNeuendorf:
I do see the point. In the current state you would need N solves to get all states. This is not a huge computational problem, but certainly inelegant. But i do not like the idea of returning a list of vectors or some other array by default. The reason for this is, that you always end up with an NxN structure and you never know, which index belongs to the states. I think that in a lot of cases the user only cares about one state and having a clearly named argument for this is the best way to do it. But i think having state=None default to returning everything would be completely fine.

fjosw:
I disagree with your statement that "you never know, which index belongs to the states" as this is clear from the docstring (which I extended in the most recent commit). Similarly `scipy.linalg.eigh` also returns the eigenvectors as a two dimensional array and also there a convention is specified. I would also argue that even if one is only interested in the energy of one state one would always have to look at the energy levels of the neighboring states for consistency checks.

Following up on your suggestion I removed the `DeprecationWarning` for the `kwarg` `state` and included it in the docstring so it is still possible to obtain only one state by specifying `state=n`. Note that the two following lines give equivalent results which is now also explained in the docstring.
```python
C.GEVP(t0=2)[n]
C.GEVP(t0=2, state=n)
```

PR #101, opened 2022-05-19
bugfix: Deltas in JSON file are now relative to o.value and not o.r_values[r_name]

s-kuberski:
This fixes a bug in the JSON I/O: 

The definition of the JSON file format states that the deltas that are saved in the file are the deviations on each ensemble relative to **the mean value**. However, we did export the deviation relative to the **replica mean**. This led to a loss of information namely the mean values on the single replica, which were set to the global mean value after import.

The bug fix now ensures that the correct replica means are restored. Within the file, the deltas within one entry `'deltas'` do not sum to zero anymore, but to `r_value - value`. The sum of all deltas within one ensemble still sums up to zero. This could be used as a cross-check. I am thinking about how to implement this, together with a speedup of the import.

Old files that were written with the bug are still read correctly, however, the r_values are not restored. New files will differ from the old ones but will respect the definition of the format.

Together with the bug fix, a small change has been made to speed up the creation of `rd['deltas']`, using numpy arrays.

### DIFF
--- a/pyerrors/input/json.py
+++ b/pyerrors/input/json.py
@@ -33,6 +33,7 @@ def create_json_string(ol, description='', indent=1):
 
     def _gen_data_d_from_list(ol):
         dl = []
+        No = len(ol)
         for name in ol[0].mc_names:
             ed = {}
             ed['id'] = name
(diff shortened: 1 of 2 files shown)


fjosw:
Thanks! I guess this change should also be reflected in the `json_schema`?

s-kuberski:
BTW: I guess that https://github.com/fjosw/ADjson.jl/blob/main/src/ADjson.jl would probably also have to be adapted...

fjosw:
If you think the current schema is already consistent I don't see a problem with it. I will take a look at the Julia version later!

PR #102, opened 2022-05-19
Feature/covariance idl2

fjosw:
The changes in this PR are subset of the changes in PR #90 plus a few additional tests.
With this patch the estimate for the correlation between two observables is always based on the intersection of samples the two are based on. In contrast to PR #90 these intersections are not computed globally for the full matrix but entry by entry which does not ensure positive definiteness. For this reason users have to be careful when estimating covariance matrices for observables which are not defined on a consistent set of samples.


fjosw:
Any objections or can I merge this PR @s-kuberski ?

PR #103, opened 2022-05-19
Bug in json format value assignment

fjosw:
I think I found another bug which only appears in the reconstruction of observables defined on multiple replica for which the average of the `r_values` is **not** equal to the `value`. In these cases the correct value was not reconstructed but instead the average of the `r_values` was assigned.


s-kuberski:
Very good (or not?), thanks for spotting this and for increasing the coverage of the tests. I hope that this resolves all issues with the I/O.

PR #104, opened 2022-05-25
Prefit correlated fits

fjosw:
When attempting a correlated fit I propose to always do an uncorrelated fit first in order to guess appropriate starting values.

### DIFF
--- a/pyerrors/fits.py
+++ b/pyerrors/fits.py
@@ -482,15 +482,15 @@ def _standard_fit(x, y, func, silent=False, **kwargs):
         chol_inv = np.linalg.inv(chol)
         chol_inv = np.dot(chol_inv, covdiag)
 
-        def chisqfunc(p):
+        def chisqfunc_corr(p):
             model = func(p, x)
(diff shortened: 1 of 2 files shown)


s-kuberski:
I think that this is a good idea. However, the argument `initial_guess` now seems to be silently ignored. Therefore, I propose to do the pre-fit only when `initial_guess` is None.

fjosw:
In the current version `initial_guess` is not ignored but passed as initial parameters to the uncorrelated fit. I feel like this is the most reliable procedure as I would expect the outcome of the uncorrelated fit to be closer to the correlated fit than any guess. The question is whether there is a situation in which the outcome of the uncorrelated fit is not sufficient for the correlated fit to converge. If one has to fine tune the guess in such a way there might be a problem with the minimization problem in the first place.

I am happy to add another option if this is needed in practice.

s-kuberski:
Ah yes, you are right, I did not see this. I cannot imagine a case where the uncorrelated fit produces results which depart too much from the true values, compared to any guess. Therefore, your implementation seems to be fine.
Maybe a slight adaption of the docstring of `total_least_squares` that explains the behavior could help to make transparent what is done.

PR #105, opened 2022-05-26
feature: fit range scan

fjosw:
This is a first version of a new method for the class `Corr` which scans fitranges for a given fit function and visualizes the dependence on the start and end time of the fit. Here is an example output:
output

### DIFF
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -3,6 +3,8 @@ from itertools import permutations
 import numpy as np
 import autograd.numpy as anp
 import matplotlib.pyplot as plt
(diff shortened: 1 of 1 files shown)


s-kuberski:
I think that this is a quite useful routine. It would be good to return `results` such that the user is able to investigate some thinks more closely. A possibility to restrict the plots to one/some fit parameters could also be helpful.

fjosw:
I fully agree, I think I returned the result of the scan in an earlier version. In the current proposal the result of the scan is stored in a nested dict, maybe we could rethink the data structure. I am also not perfectly happy with the way in which one can specify the desired scan range.

Issue #111, opened 2022-06-15
Nested loop with same variable in input.dobs


fjosw:
In lines 776 and 783 of https://github.com/fjosw/pyerrors/blob/develop/pyerrors/input/dobs.py loops with loop variable `name` are started which is already used in the higher level loop starting in line 730. I think the functionality is not affected by this but I would still propose to rename the loop variables.

PR #112, opened 2022-06-19
Feature/dobs

### LINKED ISSUE #110: Dublication in input.dobs.read_dobs
In the function `read_dobs` `content` is updated twice, once within and once after the if clause. I think lines 590 & 591 in https://github.com/fjosw/pyerrors/blob/develop/pyerrors/input/dobs.py should be removed as it is not possible to read an unzipped xml in the current version.

s-kuberski:
This closes #110 and #111. More importantly, a bug in writing the replica values (the same one as for the JSON files) has been fixed.


fjosw (approved):
Looks good, thanks!

PR #113, opened 2022-06-30
Pandas DataFrame support

fjosw:
I added export and import functionality for pandas `DataFrames` with `Obs` type columns to efficiently handle large data sets. On my local machine writing and reading O(8000) `Obs` with 9 columns of metadata (strings, ints and floats) takes less than a second each.

The serialization of the `Obs` objects is based on our standardized json format. The DataFrame is stored as a (gzipped) csv file on disk. Storage as a (gzipped) json file would work equally well but I figured that the additional metadata the json file would contain is irrelevant in this context.

### DIFF
--- a/pyerrors/input/__init__.py
+++ b/pyerrors/input/__init__.py
@@ -10,4 +10,5 @@ from . import hadrons
 from . import json
 from . import misc
 from . import openQCD
+from . import pandas
 from . import sfcf
--- a/pyerrors/input/pandas.py
+++ b/pyerrors/input/pandas.py
@@ -1,0 +1,72 @@
+import warnings
+import gzip
+import pandas as pd
+from ..obs import Obs
+from .json import create_json_string, import_json_string
+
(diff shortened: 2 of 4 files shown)


s-kuberski:
Out of curiosity: Can you explain briefly how you use this in you analyses? Instead of (potentially nested) dicts (as I use them) you keep your data in an MxN pandas dataframe which contains the Obs as well as the corresponding meta data? And when you export this to disk, you convert each Obs-like entry to a json-string which is then used to export the whole thing to csv?

This makes sense to me, although it adds another layer of information around the json format, while the possibility to add additional information to the Obs itself (description parameter of the JSON export) is not used. 

Is this faster or more memory-efficient than exporting the bunch of observables to a JSON file? I had the impression that, with the new JSON library, the process is dominated by the conversion of Obs to str.

The `auto-gamma` feature could be added to the JSON routines, as well.

PR #114, opened 2022-07-01
hash method for Obs objects

fjosw:
This is my first attempt at a reproducible hash function for `Obs` objects. I construct a tuple out of the value, the deltas, the errsq of the covobs and the names and hash that. All numerical values are cast to single precision via numpy which hopefully takes care of any rounding errors that can occur.

### DIFF
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -692,7 +692,7 @@ class Obs:
         hash_tuple += tuple([o.encode() for o in self.names])
         m = hashlib.md5()
         [m.update(o) for o in hash_tuple]
-        return int(m.hexdigest(), 16)
+        return int(m.hexdigest(), 16) & 0xFFFFFFFF
 
     # Overload comparisons
     def __lt__(self, other):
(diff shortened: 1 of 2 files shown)


s-kuberski:
The use of single precision is very smart, I had the fear that our non-binary I/O would lead to problems in this application. At some point, we could include the hash in the JSON file if the computation is fast, compared to the I/O.

fjosw:
Thanks. I tried to write the hash method in such a way that two objects that compare as equal also have the same hash but I am sure that one can construct special cases for which this is not the case. Do you think it is worth adding the hash method to the library?

The slight problem with writing the hash to a file is that this hash can probably not be easily reproduced in other programming languages. We could still add it to files as an optional check. Or we might consider a more portable hash method.

fjosw:
I followed up on my earlier comment and changed the hashing algorithm to md5 which should be portable and reproducible.

s-kuberski:
The change to md5 makes it definitely easier to compute the hash in other implementations, as well. I am not sure if it is necessary to add this to the I/O. In my opinion it crucially depends on the overhead that is introduced when writing and reading a large number of observables.

PR #116, opened 2022-07-11
Extraction of the GF coupling 

fjosw:
I added a function that can extract the gradient flow coupling from gfms files created by sfqcd.
I tried to reuse existing code as much as possible but had to make a few ugly adjustments. The whole submodule could use some refactoring but I think my patch adds the desired functionality without breaking anything or duplicating code. Please take a close look at my proposed changes.

### DIFF
--- a/pyerrors/input/openQCD.py
+++ b/pyerrors/input/openQCD.py
@@ -562,6 +562,113 @@ def read_qtop(path, prefix, c, dtr_cnfg=1, version="openQCD", **kwargs):
     integer_charge : bool
         If True, the charge is rounded towards the nearest integer on each config.
     """
+
+    return _read_flow_obs(path, prefix, c, dtr_cnfg=dtr_cnfg, version=version, obspos=0, **kwargs)
+
+
+def read_gf_coupling(path, prefix, c, dtr_cnfg=1, Zeuthen_flow=True, **kwargs):
+    """Read the gradient flow coupling based on sfqcd gradient flow measurements. See 1607.06423 for details.
+
+    Parameters
+    ----------
+    path : str
+        path of the measurement files
+    prefix : str
+        prefix of the measurement files, e.g. <prefix>_id0_r0.ms.dat.
+        Ignored if file names are passed explicitly via keyword files.
+    c : double
+        Smearing radius in units of the lattice extent, c = sqrt(8 t0) / L.
+    dtr_cnfg : int
+        (optional) parameter that specifies the number of measurements
+        between two configs.
+        If it is not set, the distance between two measurements
+        in the file is assumed to be the distance between two configurations.
+    steps : int
+        (optional) Distance between two configurations in units of trajectories /
+         cycles. Assumed to be the distance between two measurements * dtr_cnfg if not given
+    r_start : list
+        list which contains the first config to be read for each replicum.
+    r_stop : list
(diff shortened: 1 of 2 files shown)


s-kuberski:
This is a bit tricky. I have not gone into the details of the code yet, but two things certainly have to be adjusted:
- With the hardcoded normalization, the only coupling that you can compute is the one for T=L, c=0.3, Zeuthen-flow, as specified in the corresponding paper. Therefore, an exception has to be thrown for all other combinations.
- The projection on topological sector zero is strictly speaking part of the definition of the gradient flow coupling the ALPHA-SF context. Therefore, one could think to directly include it.

fjosw:
Thanks for your comments. I fully agree with the first point and will add the corresponding exceptions. I think I would prefer to do the projection in a separate step because of my current workflow but I will think about it.

pyerrors/input/openQCD.py
```diff
@@ -618,16 +730,14 @@ def read_qtop(path, prefix, c, dtr_cnfg=1, version="openQCD", **kwargs):
     r_stop_index = []
     deltas = []
     configlist = []
+    if not zeuthen:
+        obspos += 8
```
s-kuberski, line 734: Is this correct for all file and code versions? I think it might only be true for the `sfqcd` files. At least this is how I had implemented it in lines 627-630 of the old code version.

s-kuberski:
The current version is definitely better. You could enforce T=L by having explicit keywords for this. This would also help in a future version where others norms are implemented and it would also allow you to remove the parsing of the tag of the plaquette. Within `_read_flow_obs', the value of L would be checked against the one in the file, if you pass it.
I have added a comment regarding `obspos`.

pyerrors/input/openQCD.py
fjosw: I think the variable `obspos` is only relevant for `sfqcd` files. For the `openQCD` files the function currently always extracts the topological charge based on the gradient flow regardless of what is specified for `obspos` or `Zeuthen_flow`.

s-kuberski: Yes, you are right, I did not see that.

fjosw:
I now added a check for `T=L` by passing both `T` and `L` via the tag. I feel like this is the least disruptive way and does not require additional keywords and complexity (of which we already have a lot in this module). What do you think?

s-kuberski:
I think that this is 

> I now added a check for `T=L` by passing both `T` and `L` via the tag. I feel like this is the least disruptive way and does not require additional keywords and complexity (of which we already have a lot in this module). What do you think?

This is a good solution. In case one implements any other norm, the correct one can then be chosen based on the content of the file.

s-kuberski:
One check is still missing: The norm is computed for the Zeuthen flow, so `Zeuthen_flow` has to be `True`. You could set it to `True` by default. The way it works right now, as far as I can see, the default value is `False` and this is not caught such that the result for the coupling is wrong. This also excludes all non-sfqcd files, because they don't have the Zeuthen flow.

fjosw:
I think the Zeuthen flow was already default for the coupling but I now added an explicit Exception if one attempts to use the gradient flow observables to compute the coupling.

s-kuberski:
I have checked the code against my own implementation and did not find any issues.

fjosw:
Thanks for double checking!

PR #117, opened 2022-07-19
is_matrix_symmetric

fjosw:
I implemented a new method `is_matrix_symmetric` which uses the new hash function to efficiently check whether a correlator matrix is symmetric. This check is now performed within `matrix_symmetric` and speeds up the method in case the matrix is already symmetric. This in turn accelerates the GEVP method which always calls `matrix_symmetric`.

### DIFF
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -242,6 +242,8 @@ class Corr:
         if self.N == 1:
             raise Exception("Only works for correlator matrices.")
         for t in range(self.T):
+            if self[t] is None:
+                continue
             for i in range(self.N):
                 for j in range(i + 1, self.N):
                     if self[t][i, j] is self[t][j, i]:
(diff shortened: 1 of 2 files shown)


s-kuberski:
I did not have in mind that we symmetrize explicitly in the GEVP. To me, it makes a lot of sense to circumvent doing this, when the matrix is already symmetric. Do you have an idea how much faster it is to do the check instead of performing an explicit symmetrization?
The test could in principle be performed based on a double-comparison of mean values, right? I guess that it is highly unlikely that two mean values are equal to double precision when the Obs differ. Of course, the hash-based comparison more or less guarantees equality, but if it adds a lot of work, one could get rid of it.

fjosw:
The timings of course depend on the exact correlator matrix but it looks to me as if the hash comparison is always subleading. Checking and then assigning by reference (as in the GEVP method) should be about 10x as fast. In `matrix_symmetric` I don't return by reference but construct a copy via `1.0 * self`. This step consumes most of the time but is still about twice as fast as explicitly computing the symmetrization which requires both an addition and a multiplication.

Checking only the mean values would be about 100x faster but one could construct edge cases in which the method fails.

PR #121, opened 2022-10-05
Least square fit with numerical differentiation

fjosw:
I implemented the option to perform the error propagation for least squares fits with numerical differentiation instead of automatic differentiation. Using this option is ~~considerably~~ slower but can deal with fit functions which contain special functions or which are not analytically known.

### DIFF
--- a/pyerrors/fits.py
+++ b/pyerrors/fits.py
@@ -10,8 +10,10 @@ from matplotlib import gridspec
 from scipy.odr import ODR, Model, RealData
 import iminuit
 from autograd import jacobian as auto_jacobian
+from autograd import hessian as auto_hessian
 from autograd import elementwise_grad as egrad
 from numdifftools import Jacobian as num_jacobian
+from numdifftools import Hessian as num_hessian
 from .obs import Obs, derived_observable, covariance, cov_Obs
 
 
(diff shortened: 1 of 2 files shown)


pyerrors/fits.py
```diff
@@ -406,14 +418,19 @@ def chisqfunc(p):
     if not m.fmin.is_valid:
         raise Exception('The minimization procedure did not converge.')
 
-    hess_inv = np.linalg.pinv(jacobian(jacobian(chisqfunc))(params))
+    hess = jacobian(jacobian(chisqfunc))(params)
```
s-kuberski, line 421: You could use the `hessian()` here as well, right?

fjosw: Yes, for now I only adapted the `_standard_fit` function, but I can do that in the next iteration.

s-kuberski:
I have not yet tested this feature but I very much like the possibility to be more flexible with the whole package in general and the fitting routines to make it easier to port existing analyses or to make it possible to fit non-analytic functions. 
We could add a small example how to define special functions with known analytic derivatives for a use in autograd along the lines of what is shown in https://github.com/HIPS/autograd/blob/master/docs/tutorial.md#extend-autograd-by-defining-your-own-primitives. (This would be orthogonal to the new feature)

fjosw:
> I have not yet tested this feature but I very much like the possibility to be more flexible with the whole package in general and the fitting routines to make it easier to port existing analyses or to make it possible to fit non-analytic functions. We could add a small example how to define special functions with known analytic derivatives for a use in autograd along the lines of what is shown in https://github.com/HIPS/autograd/blob/master/docs/tutorial.md#extend-autograd-by-defining-your-own-primitives. (This would be orthogonal to the new feature)

That is a good idea. Would you add the example in the docstring, in the documentation or as a separate notebook?

s-kuberski:
I am not sure. If you put it in some docstring, it might not be found by anyone who needs this. One could add it to one of the notebooks in the examples or add it to the documentation, I think.

Issue #123, opened 2022-10-14
Automatic windowing method fails for gapped and irregular chains


s-kuberski:
The automatic windowing procedure fails in cases where the Monte Carlo history is gapped, e.g., with stride 2, and on top of this contains also holes. The main reason for this behavior is that we do not expand the chain if `idl` is a `range`, but we do so, if it is a `list`. Minimal example, close to what is done in `test_gamma_method_irregular`:

```python
N = 10000
def gen_autocorrelated_array(inarr, rho):
    outarr = np.copy(inarr)
    for i in range(1, len(outarr)):
        outarr[i] = rho * outarr[i - 1] + np.sqrt(1 - rho**2) * outarr[i]
    return outarr

arr = np.random.normal(1, .2, size=N)
carr = gen_autocorrelated_array(arr, .8)
a = pe.Obs([carr], ['a'])
a.gamma_method()

arrt = [carr[i] for i in range(len(carr)) if i % 2 == 1]
idlt = [i for i in range(len(carr)) if i % 2 == 1]
for el in [int(e) for e in N * np.random.uniform(size=int(N/4))]:
    arrt = arrt[:el] + arrt[el + 1:]
    idlt = idlt[:el] + idlt[el + 1:]
at = pe.Obs([arrt], ['a'], idl=[idlt])
at.gamma_method(tau_exp=2)
at.plot_rho()
```

There are probably multiple ways to fix this behavior and I am not yet certain which one is the best. There is the possibility to determine the regular gaps based on the rho that we have calcuated via
```python
# detect holes:
            gapsize = 1
            if not all(isinstance(self.idl[r_name], range) for r_name in e_content[e_name]):
                for n in range(1, w_max):
                    if abs(self.e_rho[e_name][n]) < 1.e-13:
                        gapsize += 1
                    else:
                        break
```
before performing the automatic windowing procedure. This one would then be done via 
```python
                    # Standard automatic windowing procedure
                    tau = self.S[e_name] / np.log((2 * self.e_n_tauint[e_name][1:] + 1) / (2 * self.e_n_tauint[e_name][1:] - 1)) * gapsize
                    g_w = np.exp(- np.arange(1, w_max) / tau) - tau / np.sqrt(np.arange(1, w_max) * e_N)
                    for n in range(gapsize, w_max, gapsize):
                        if n < w_max // 2 - 2:
                            _compute_drho(n + 1)
                        if (g_w[n - gapsize] < 0 or n >= w_max - 1):
[...]

PR #124, opened 2022-10-17
fix: bug in automatic window for irregular chains fixed

s-kuberski:
This addresses the issue and discussion in #123. 
A few comments:

- `gapsize` is determined from the first replica, assuming that it is the same for all of them. A check could be done upon initialization of the (derived) obs. I had the feeling that it should not be done when the gamma method is called.
- The fix for the critical slowing down analysis seems to be rather trivial. I hope that everything is done correclty here. I have not yet implemented a test for this feature, this should probably be done. A general issue is in handling tau_exp. We could use `tau_exp` in units of `configs` instead of the step size in the case of regular, gapped MC chains. This could avoid a lot of confusion.
- We have to decide how to display the integrated autocorrelation times (and windows). The standard was that this is in units of the stepsize for regular chains and in units of 1 for irregular chains. This is not the case for the fix presented here: The integrated autocorrelation time is given in terms of the `gapsize`. This is confusing. 
We could scale the output of `details` to consistently print the time in the same units. We should probably print the regular step size of the chain, i.e., gapsize in the case of irregular chains. Otherwise, regardless of how we do things, it can be very confusing to determine the integrated autocorrelation time in units of, say, MDU.

### DIFF
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -305,7 +305,7 @@ class Obs:
                         self.e_dtauint[e_name] = np.sqrt(self.e_n_dtauint[e_name][n] ** 2 + texp ** 2 * self.e_drho[e_name][n + 1] ** 2)
                         # Error of tau_exp neglected so far, missing term: self.e_rho[e_name][n + 1] ** 2 * d_tau_exp ** 2
                         self.e_dvalue[e_name] = np.sqrt(2 * self.e_tauint[e_name] * e_gamma[e_name][0] * (1 + 1 / e_N) / e_N)
-                        self.e_ddvalue[e_name] = self.e_dvalue[e_name] * np.sqrt((n + 0.5) / e_N)
+                        self.e_ddvalue[e_name] = self.e_dvalue[e_name] * np.sqrt((n / gapsize + 0.5) / e_N)
                         self.e_windowsize[e_name] = n
                         break
             else:
@@ -327,7 +327,7 @@ class Obs:
                             self.e_tauint[e_name] = self.e_n_tauint[e_name][n] * (1 + (2 * n / gapsize + 1) / e_N) / (1 + 1 / e_N)  # Bias correction hep-lat/0306017 eq. (49)
                             self.e_dtauint[e_name] = self.e_n_dtauint[e_name][n]
                             self.e_dvalue[e_name] = np.sqrt(2 * self.e_tauint[e_name] * e_gamma[e_name][0] * (1 + 1 / e_N) / e_N)
-                            self.e_ddvalue[e_name] = self.e_dvalue[e_name] * np.sqrt((n + 0.5) / e_N)
+                            self.e_ddvalue[e_name] = self.e_dvalue[e_name] * np.sqrt((n / gapsize + 0.5) / e_N)
                             self.e_windowsize[e_name] = n
                             break
 
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -279,11 +279,17 @@ class Obs:
                 tmp = self.e_rho[e_name][i + 1:w_max] + np.concatenate([self.e_rho[e_name][i - 1::-1], self.e_rho[e_name][1:w_max - 2 * i]]) - 2 * self.e_rho[e_name][i] * self.e_rho[e_name][1:w_max - i]
                 self.e_drho[e_name][i] = np.sqrt(np.sum(tmp ** 2) / e_N)
 
-            # detect regular step size in irregular MC chain
-            gapsize = 1
-            for r_name in e_content[e_name][:1]:
-                if not isinstance(self.idl[r_name], range):
-                    gapsize = np.min(np.diff(self.idl[r_name]))
+            gaps = []
+            for r_name in e_content[e_name]:
+                if isinstance(self.idl[r_name], range):
+                    gaps.append(1)
+                else:
+                    gaps.append(np.min(np.diff(self.idl[r_name])))
+
+            if not np.all([gi == gaps[0] for gi in gaps]):
+                raise Exception(f"Replica for ensemble {e_name} are not equally spaced.", gaps)
+            else:
+                gapsize = gaps[0]
 
             _compute_drho(gapsize)
             if self.tau_exp[e_name] > 0:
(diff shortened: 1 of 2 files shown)


pyerrors/obs.py
```diff
@@ -305,12 +311,13 @@ def _compute_drho(i):
...
-                    g_w = np.exp(- np.arange(1, w_max) / tau) - tau / np.sqrt(np.arange(1, w_max) * e_N)
+                    tau = self.S[e_name] / np.log((2 * self.e_n_tauint[e_name][gapsize::gapsize] + 1) / (2 * self.e_n_tauint[e_name][gapsize::gapsize] - 1))
+                    g_w = np.exp(- np.arange(1, len(tau) + 1) / tau) - tau / np.sqrt(np.arange(1, len(tau) + 1) * e_N)
                     for n in range(1, w_max):
                         if n < w_max // 2 - 2:
-                            _compute_drho(n + 1)
+                            _compute_drho(gapsize * n + gapsize)
                         if g_w[n - 1] < 0 or n >= w_max - 1:
+                            n *= gapsize
                             self.e_tauint[e_name] = self.e_n_tauint[e_name][n] * (1 + (2 * n + 1) / e_N) / (1 + 1 / e_N)  # Bias correction hep-lat/0306017 eq. (49)
```
fjosw, line 321: I think the bias correction should also be modified by the `gapsize`. So instead of `n` one should divide by `n \ gapsize`.

I wrote a short test that would check this:
```python
N = 1000
def gen_autocorrelated_array(inarr, rho):
    outarr = np.copy(inarr)
    for i in range(1, len(outarr)):
        outarr[i] = rho * outarr[i - 1] + np.sqrt(1 - rho**2) * outarr[i]
    return outarr


dat = gen_autocorrelated_array(np.random.normal(1, .2, size=N), .8)

idl_a = list(range(0, 1001, 1))
idl_a.remove(101)

oa = pe.Obs([dat], ["ens1"], idl=[idl_a])
oa.gamma_method()
tau_a = oa.e_tauint["ens1"]

idl_b = list(range(0, 10010, 10))
idl_b.remove(1010)

ob = pe.Obs([dat], ["ens1"], idl=[idl_b])
ob.gamma_method()
tau_b = ob.e_tauint["ens1"]

assert np.isclose(tau_a, tau_b)
```

pyerrors/obs.py
```diff
@@ -279,14 +279,20 @@ def _compute_drho(i):
                 tmp = self.e_rho[e_name][i + 1:w_max] + np.concatenate([self.e_rho[e_name][i - 1::-1], self.e_rho[e_name][1:w_max - 2 * i]]) - 2 * self.e_rho[e_name][i] * self.e_rho[e_name][1:w_max - i]
                 self.e_drho[e_name][i] = np.sqrt(np.sum(tmp ** 2) / e_N)
 
-            _compute_drho(1)
+            # detect regular step size in irregular MC chain
+            gapsize = 1
```
fjosw, line 283: The combination of `np.min` and `np.diff` seems to be pretty fast even for large data sets. We could consider explicitly checking that gaps are consistent across replica

```python
gaps = []
for r_name in e_content[e_name]:
    if isinstance(self.idl[r_name], range):
        gaps.append(1)
    else:
        gaps.append(np.min(np.diff(self.idl[r_name])))

if not np.all(gaps == gaps[0]):
    raise Exception(f"Replica for ensemble {e_name} are not equally spaced.")
else:
    gapsize = gaps[0]
```
For two replica with `1000` samples each the overhead on my machine is 76.6 µs ± 2.17 µs compared to 34.7 µs ± 1.04 µs for your variant. The full `gamma_method` takes 11 ms ± 338 µs.

fjosw (changes requested):

fjosw:
The line numbers of my comments are somehow off, I hope they are still understandable.

pyerrors/obs.py
s-kuberski: That is indeed very fast, I'll include it.

pyerrors/obs.py
s-kuberski: Ah yes, you are right. I had overlooked this one. It also appears in the critical slowing down analysis.

pyerrors/obs.py
```diff
@@ -305,13 +317,14 @@ def _compute_drho(i):
...
                         if n < w_max // 2 - 2:
-                            _compute_drho(n + 1)
+                            _compute_drho(gapsize * n + gapsize)
                         if g_w[n - 1] < 0 or n >= w_max - 1:
-                            self.e_tauint[e_name] = self.e_n_tauint[e_name][n] * (1 + (2 * n + 1) / e_N) / (1 + 1 / e_N)  # Bias correction hep-lat/0306017 eq. (49)
+                            n *= gapsize
+                            self.e_tauint[e_name] = self.e_n_tauint[e_name][n] * (1 + (2 * n / gapsize + 1) / e_N) / (1 + 1 / e_N)  # Bias correction hep-lat/0306017 eq. (49)
                             self.e_dtauint[e_name] = self.e_n_dtauint[e_name][n]
                             self.e_dvalue[e_name] = np.sqrt(2 * self.e_tauint[e_name] * e_gamma[e_name][0] * (1 + 1 / e_N) / e_N)
                             self.e_ddvalue[e_name] = self.e_dvalue[e_name] * np.sqrt((n + 0.5) / e_N)
```
fjosw, line 330: I think the bias correction also has to be adjusted for the `ddvalue` entries.

fjosw (approved):

fjosw:
Should we discuss the modification of `details` here or in a separate PR?

fjosw:
I think I would prefer to keep the output as it is but add information about the step/gapsize.
This would give a similar estimate for data defined on
- every *n*th configuration
- every *n*th configuration but the *m*th
If we rescale these two cases would result in very different estimates for tau_int.

The question is how to convey the relevant information.
- in units of {gapsize}
- in steps of {gapsize} configurations
- ...

fjosw:
In case we want to rescale the integrated autocorrelation time with the gapsize $s>1$ the relevant relation should be
$$\tau_\mathrm{int}(s)=0.5+\frac{1}{\mathrm{e}^{s/\tau_\mathrm{int}(1)}-1}$$
based on the assumption that $\rho$ is a single exponential function.

fjosw:
Maybe `in units of 2 configs` is a bit shorter? But we could certainly lower the number of digits displayed. I guess for t_int there is no need for more than 3 or 4 digits or what do you think?

s-kuberski:
Your proposal for the string, together with a reduction of digits, seems to be a good combination!

PR #125, opened 2022-10-19
Improved details output

fjosw:
I reworked the output of the `details` method based on the discussion in #124.
In addition to that I added greek unicode characters where applicable, changed the output of tau_int with its error to paranthesis notation and adjusted the right alignment of the information about the windowing procedure.

The output for an observable with support on a single ensemble looks like this:
```
Result	 1.38625189e+01 +/- 2.95299110e-01 +/- 8.34709829e-02 (2.130%)
 τ_int	 121(51) in units of 1 config         	(S=2.00)
5000 samples in 1 ensemble:
  · Ensemble 'B20' : 5000 configurations (from 1 to 5000)
```

and for multiple ensembles like this:
```
Result	 1.30947356e+01 +/- 1.17152209e+00 +/- 3.83826885e-01 (8.947%)
 Ensemble errors:
 A10 	 1.130136e+00 +/- 3.971589e-01
 τ_int	 119(36) in units of 1 config         	(τ_exp=160.00, N_σ=1)
 H65 	 3.086357e-01 +/- 8.783925e-02
 τ_int	 12.1(5.1) in units of 100000 configs 	(S=2.00)
1500 samples in 2 ensembles:
  · Ensemble 'A10' 
    · Replicum 'r1' : 500 configurations (from 1 to 500)
    · Replicum 'r2' : 500 configurations (from 1 to 500)
  · Ensemble 'H65' : 500 configurations (from 1 to 49900001 in steps of 100000)
```

### DIFF
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -433,7 +433,7 @@ class Obs:
                         if isinstance(self.idl[value[0]], range):
                             my_string += f' (from {self.idl[value[0]].start} to {self.idl[value[0]][-1]}' + int(self.idl[value[0]].step != 1) * f' in steps of {self.idl[value[0]].step}' + ')'
                         else:
-                            my_string += ' (irregular range)'
+                            my_string += f' (irregular range from {self.idl[value[0]][0]} to {self.idl[value[0]][-1]})'
                     else:
                         sublist = []
                         for v in value:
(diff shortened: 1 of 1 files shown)


s-kuberski:
Looks very good, I like all of the changes! I think that this will help to avoid errors in the interpretation of the integrated autocorrelation time. 

One could think of displaying `(irregular range from {self.idl[v].start} to {self.idl[v][-1]}')` instead of `(irregular range)`, just because you started to adapt the output and this could add useful information

PR #126, opened 2022-10-19
Fix: gapped irregular dtauint

fjosw:
This PR fixes the calculation of `dtauint` for gapped irregular chains which appeared in connection with #124. The missing part was to divide `np.abs(np.arange(w_max)` by `gapsize`. I also added a test which checks this.


s-kuberski (approved):
Ah, another thing that I have overlooked. Thanks for fixing it!

PR #127, opened 2022-10-19
feat: added log-derivatives and symmetric effective mass

s-kuberski:
The log-derivatives are defined as 
$\partial_t C(t) = C(t) \cdot \partial_t \log[C(t)] $
and
$\partial_t^2 C(t) = C(t) [\partial_t^2 \log[C(t)] + ( \partial_t \log[C(t)])^2] $
and they are exact for single-exponential functions, i.e., they do not induce any discretization error. Of course, the user could implement these themselves, but I find this way more convenient.

I have also added the log-effective mass based on a symmetric derivative.

### DIFF
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -523,7 +523,7 @@ class Corr:
         ----------
         variant : str
             decides which definition of the finite differences derivative is used.
-            Available choice: symmetric, forward, backward, improved, default: symmetric
+            Available choice: symmetric, forward, backward, improved, log, default: symmetric
         """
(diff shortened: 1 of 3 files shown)


fjosw:
Looks very good, thanks! The tests seem to fail because of unfortunate random numbers. One quick fix would be to slightly increase the tolerance in `test_fit_corr_independent`.

s-kuberski:
You mean it was just bad luck that it hit my commit? Too bad... I'll adapt the bound.

fjosw (approved):

PR #128, opened 2022-10-21
feat: Speed up covariance for irregular MC chains

s-kuberski:
This PR speeds up the computation of covariance by orders of magnitude, when irregular MC chains are involved.

I removed `_collapse_deltas_for_merge` as it did the same job as `_reduce_deltas`. I also improved the implementation of `_reduce_deltas`, mainly by changing two things:

- I now use `groupby` to check if to `idx` are equal. For covariance matrices of correlators, this means that no reduction has to be performed, at all. In principle, this is already checked in `_intersection_idx` within `_covariance_element` such that we could do some rewriting: Export a boolean that tells you if all lists are equal when calling `_intersection_idx` and use this to skip the call for the reduction in `calc_gamma`. However, I have the impression that the comparison using groupby is very efficient (less than 1µs for realistic cases) such that it is okay to check this anyways. In this way, `reweight` should also profit from the speedup.
- I removed my manual matching of the `idx` to select the correct deltas by using `numpy.intersect1d`. I think that this should be the most efficient way to do it and I see a significant speedup w.r.t. the old implementation and even more so w.r.t `_collapse_deltas_for_merge`.

### DIFF
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -1097,34 +1097,6 @@ def _expand_deltas_for_merge(deltas, idx, shape, new_idx):
     return np.array([ret[new_idx[i] - new_idx[0]] for i in range(len(new_idx))])
 
 
-def _collapse_deltas_for_merge(deltas, idx, shape, new_idx):
-    """Collapse deltas defined on idx to the list of configs that is defined by new_idx.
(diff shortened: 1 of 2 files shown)


fjosw (approved):
Looks good!

PR #130, opened 2022-10-25
feat: JSON export can handle numpy.int64 entries

s-kuberski:
This change handles cases where entries of the data structure that is to be exported to a JSON file, e.g., via `pe.input.json.dump_dict_to_json` contains a `numpy.int64`. Before, an error has been thrown, because the `dunps` function now gets a plain `int` via the type conversion. If any more such cases exist, they could be included in the new function `_jsonifier'.

### DIFF
--- a/pyerrors/input/json.py
+++ b/pyerrors/input/json.py
@@ -186,10 +186,15 @@ def create_json_string(ol, description='', indent=1):
         else:
             raise Exception("Unkown datatype.")
 
+    def _jsonifier(o):
+        if isinstance(o, np.int64):
+            return int(o)
+        raise TypeError('%r is not JSON serializable' % o)
+
     if indent:
-        return json.dumps(d, indent=indent, ensure_ascii=False, write_mode=json.WM_SINGLE_LINE_ARRAY)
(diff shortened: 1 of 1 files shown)


fjosw:
I think I found the problem the CI is facing: Today `pytest` 7.2 was released which no longer depends on the library `py` (https://docs.pytest.org/en/stable/changelog.html). `pytest-benchmark`, however, still implicitly does. Including `pip install py` in the CI workflow should fix this.

fjosw:
I updated the pytest workflow on the `develop` branch.

fjosw:
Should I merge this or do you think there are other problematic conversions that need a closer look?

s-kuberski:
There is something else that is connected to the export of dictionaries (or the case where a dictionary is saved as description of some other data) that I was not able to investigate, yet: The JSON standard requires the keys of these dictionaries to be strings. The standard ```json``` package converts strings silently to string on export. So the export works but after import, your dictionary is not the same anymore. This was the standard behavior in the past.
The ```rapidjson``` package does not do this implicit conversion but throws an error instead. It is a bit a matter of design what to do in our case. I am a bit in favor of having a silent conversion (and adding this as specification to the exporting functions).

fjosw:
I experienced this bug in production so I decided to merge the fix. The other issue could go into a separate PR.

PR #134, opened 2022-12-16
sample implementation of a (uncorrelated) combined fit

PiaLJP:
A combined fit is performed when dictionaries (instead of lists) are handed over to the fits.least_squares function.


fjosw:
Hi Pia, it looks like your PR breaks a few tests. Please have a look at CONTRIBUTING.md where I tried to explain how to run the tests on a local machine.

s-kuberski:
Thanks for the implementation. Before I inspect everything in detail, I'd like to ask a general question: Is it really necessary to do the combined fit in a new routine that has many common aspects with `_standard_fit`? I have the feeling that this will increase the amount of work that will be needed in the future to implement new routines and to fix and avoid bugs. 

My first attempt would have been to write a wrapper around `_standard_fit` or to adapt the first lines of `_standard_fit` such that proper combinations x, y, func are handed to the routine. For x and y this is trivial, for func one could just settle on a list of functions that is filled with the same function over and over again if no combined fit is done. Did you look at this and decided to not to do this for some reason (performance?)?

For the future of the fit routines in `pyerrors`, I'd rather merge all the fit routines (i.e., also `_prior_fit` would be considered here) instead of adding new routines.

fjosw:
I agree that this is not a good solution and results in a lot of unnecessary code duplication and maintainability problems. I see this PR as a step towards a new version of the fit module as it already defines the interface for combined fits and provides test cases which will also have to be fulfilled by the final version which could be very helpful in the next development phase.

In this spirit it would be very interesting to hear your opinion on the interface @PiaLJP designed and the overall functionality, not so much the actual implementation. As soon as we agree on the interface we could then start to rewrite and refactor the code.

pyerrors/fits.py
```diff
@@ -112,6 +141,8 @@ def func(a, x):
         The possible methods are the ones which can be used for scipy.optimize.minimize and
         migrad of iminuit. If no method is specified, Levenberg-Marquard is used.
         Reliable alternatives are migrad, Powell and Nelder-Mead.
+    tol: float, optional
+        can only be used for combined fits and methods other than Levenberg-Marquard
```
s-kuberski, line 145: For future reference: We should explain the impact of this parameter and it should be considered in all fits where the tolerance of the fit method may be adjusted by the user.

pyerrors/fits.py
```diff
@@ -94,9 +97,35 @@ def func(a, x):
             (x1, x2) = x
             return a[0] * x1 ** 2 + a[1] * x2
         ```
+        It is important that all numpy functions refer to autograd.numpy, otherwise the differentiation
+        will not work.
+
+    OR For a combined fit:
+
+    Do not need to use ordered dictionaries: python version >= 3.7: Dictionary order is guaranteed to be insertion order.
+    (https://docs.python.org/3/library/stdtypes.html#dict-views) Ensures that x, y and func values are mapped correctly.
```
s-kuberski, line 106: Why does the order matter when you use a dictionary? See my comment below.

pyerrors/fits.py
```diff
@@ -130,6 +161,10 @@ def func(a, x):
     '''
     if priors is not None:
         return _prior_fit(x, y, func, priors, silent=silent, **kwargs)
+
+    elif (type(x) == dict and type(y) == dict and type(func) == dict):
+        return _combined_fit(x, y, func, silent=silent, **kwargs)
+
```
s-kuberski, line 167: Fine for now, in the future we have to cover all possible cases.

pyerrors/fits.py
```diff
@@ -656,6 +690,189 @@ def chisqfunc_compact(d):
...
+    if kwargs.get('num_grad') is True:
+        jacobian = num_jacobian
+        hessian = num_hessian
+    else:
+        jacobian = auto_jacobian
+        hessian = auto_hessian
+
+    x_all = np.concatenate([np.array(o) for o in x.values()])
+    y_all = np.concatenate([np.array(o) for o in y.values()])
```
s-kuberski, line 709: You don't have to rely on the order of the dictionary here. Use the keys of one of the dictionaries to ensure that corresponding (key, value) pairs of the three dictionaries are properly matched.
``` python
keylist = list(x.keys())
x_all = np.concatenate([np.array(x[k]) for k in keylist])
y_all = np.concatenate([np.array(y[k]) for k in keylist])
```
You should also use this to check for consistency of the three dictionaries (do they contain the same keys? Do the (x, y)-pairs match for all keys? ...)

pyerrors/fits.py
```diff
@@ -656,6 +690,189 @@ def chisqfunc_compact(d):
...
+
+    y_f = [o.value for o in y_all]
+    dy_f = [o.dvalue for o in y_all]
+
+    if len(x_all.shape) > 2:
+        raise Exception('Unknown format for x values')
+
+    # number of fit parameters
+    n_parms_ls = []
+    for key in func.keys():
```
s-kuberski, line 719: Same as above. Use `keylist` or something similar.

pyerrors/fits.py
```diff
@@ -656,6 +690,189 @@ def chisqfunc_compact(d):
...
+        if not callable(func[key]):
+            raise TypeError('func (key=' + key + ') is not a function.')
+        if len(x[key]) != len(y[key]):
+            raise Exception('x and y input (key=' + key + ') do not have the same length')
+        for i in range(42):
```
s-kuberski, line 724: This is also true for the other routines, but I just thought about it: Is 42 a good upper limit for the number of fit parameters? I guess so, but if you do combined fits, the probability that this is not the case is increased.

pyerrors/fits.py
```diff
@@ -656,6 +690,189 @@ def chisqfunc_compact(d):
...
+    def chisqfunc(p):
+        chisq = 0.0
+        for key in func.keys():
+            x_array = np.asarray(x[key])
+            model = anp.array(func[key](p, x_array))
+            y_obs = y[key]
+            y_f = [o.value for o in y_obs]
+            dy_f = [o.dvalue for o in y_obs]
+            C_inv = np.diag(np.diag(np.ones((len(x_array), len(x_array))))) / dy_f / dy_f
+            chisq += anp.sum((y_f - model) @ C_inv @ (y_f - model))
```
s-kuberski, line 757: An alternative solution would be to create a list of functions a la 
``` python
func_list = np.concatenate([[func[k]] * len(x[k]) for k in keylist])
```
and to define the model via
``` python
model = anp.array([func_list[i](p, x_all[i]) for i in range(len(x_all))])
```
and then combine it with `y` as it is done in the standard routine. 
In this way, you would not have to do create all the arrays and lists over and over again. However, you don't use the feature to call the function with an array of values, which costs probably some time.

As third solution, you could ensure that the values of `x` are arrays at the beginning of the function and create dictionaries for `y_f` and `dy_f` that match `y` such that you can refer to them in this routine to ensure maximum performance.

pyerrors/fits.py
s-kuberski: This does appear a few times later in the code, where I don't highlight it.

pyerrors/fits.py
```diff
@@ -656,6 +690,189 @@ def chisqfunc_compact(d):
...
+        chisq = 0.0
+        list_tmp = []
+        c1 = 0
+        c2 = 0
+        for key in func.keys():
+            x_array = np.asarray(x[key])
+            c2 += len(x_array)
+            model = anp.array(func[key](d[:n_parms], x_array))
+            y_obs = y[key]
+            dy_f = [o.dvalue for o in y_obs]
```
s-kuberski, line 821: see my comments in the chisq function

tests/fits_test.py
```diff
@@ -608,6 +590,149 @@ def f(a, x):
...
+    for key in funcs_num.keys():
+        [item.gamma_method() for item in ys[key]]
+
+    num = pe.fits.least_squares(xs, ys, funcs_num, num_grad=True)
+    auto = pe.fits.least_squares(xs, ys, funcs_auto, num_grad=False)
+
+    assert(num[0] == auto[0])
+    assert(num[1] == auto[1])
+
+
```
s-kuberski, line 735: When you have adapted the method such that the order of the dictionaries does not matter, this should be checked here explicitly.

s-kuberski (changes requested):
Thank you very much for your effort. Let's go with the dictionary solution by now. In contrast to the other solutions that I had in mind, it would remove the need to have x, y and func in the same order, if the implementation is adapted accordingly.

pyerrors/fits.py
fjosw: This was more of a Douglas Adams reference than a meaningful upper bound. In practice, this should probably be more than enough but we could also raise it to 100 or so. I would expect the performance implications to be minimal.

pyerrors/fits.py
```diff
@@ -763,7 +768,7 @@ def chisqfunc(p):
 
     if output.method != 'Levenberg-Marquardt':
         if output.method == 'migrad':
-            tolerance = 1e-4
+            tolerance = 1e-1 # default value set by iminuit
```
fjosw, line 771: Hi Pia, I just had a quick look at your changes, thanks! Is there a reason why you changed the migrad solver tolerance? In the past I have seen problems with using `1e-1`, that's why I lowered it back then.

pyerrors/fits.py
```diff
@@ -705,8 +705,19 @@ def _combined_fit(x, y, func, silent=False, **kwargs):
...
-    y_all = np.concatenate([np.array(o) for o in y.values()])
+    key_ls = sorted(list(x.keys()))
+
+    if sorted(list(y.keys())) != key_ls:
+        raise Exception('x and y dictionaries do not contain the same keys.')
+
+    if sorted(list(func.keys())) != key_ls:
+        raise Exception('x and func dictionaries do not contain the same keys.')
+
+    if sorted(list(func.keys())) != sorted(list(y.keys())):
```
fjosw, line 716: I think this check is unnecessary. If a=b and a=c then b=c.

pyerrors/fits.py
PiaLJP: No, I just reset it.

pyerrors/fits.py
PiaLJP: Yes, I removed it, thanks!

pyerrors/fits.py
s-kuberski: We could catch the case where undefined behavior will happen, i.e., when the user does not supply two lists and a function or three dicts. This could help to identify the mistake. Otherwise, some error will be thrown in `_standard_fit` that is not necessarily well understandable.

s-kuberski:
Hi. Thanks for adapting the code. I have made one minor comment that we could resolve before merging.

s-kuberski (approved):

PR #136, opened 2023-01-02
fix: corrected small bug in qtop_projection with several replica

s-kuberski:
The projection to sectors of the topological charge was not done correctly in the case of several replica because the value of the primary observable had been extracted from the mean and not the replica mean. 

The impact of this bug is expected to be very minor since it only has an impact, if the difference between mean and replica mean leads to a shift from one sector to the other.


fjosw (approved):
Looks good, thanks!

PR #137, opened 2023-01-04
Alias gm for gamma_method

fjosw:
After typing `gamma_method` for the thousands time I thought it could be useful to have a short alias like `gm`. Any thoughts or alternative naming suggestions?


jkuhl-uni:
This sounds like a good idea. Off the top of my head I wouldn't know any better naming :)

s-kuberski:
Certainly a good idea. The name is seems to be fine.

PR #139, opened 2023-01-06
fix: remove _filter_zeroes

### LINKED ISSUE #138: Issues with _filter_zeroes and Corr
I have noticed issues with `_filter_zeroes`. These appear when two observables with vastly different variances, defined on different subsets of an ensemble, are combined (e.g. by renormalizing a noisy correlator). In this case it can happen that configurations that belong only to the observable with the small variance are filtered out by chance. This is something that I did not have in mind when designing the method.
When this does happen to elements of correlators, the export to json fails because the Obs of the Corr are not defined on the same set of configurations anymore.

One can construct a case artificially:
```python 
li = [-pe.pseudo_Obs(.2, .1, 'a', samples=10) for i in range(6)]
for i in range(len(li)):
    li[i].idl['a'] = range(1, 21, 2)
b = pe.pseudo_Obs(1, 1e-11, 'a', samples=30)
c *= b
print(c[2].idl)
print(c[3].idl)
{'a': [1, 3, 5, 7, 9, 10, 11, 13, 15, 17, 19]}
{'a': [1, 3, 4, 5, 7, 9, 10, 11, 12, 13, 14, 15, 16, 17, 19, 20, 26, 30]}
```
This is certainly a rare event but I have observed it a number of times in the past and this is quite annoying. 

A possible fix would be to remove the automatic filtering altogether and instead to put the feature in a function that may be called by the user. In this case, exact cancellations would manifest themselves as vanishing deltas and would not be filtered out anymore. I tend to think that this is better to live with than silent removals of configurations from the obs.

If we like to keep the feature, I see a number of possible improvements (use mean instead of the median in `_filter_zeroes`, lower filter_eps, ...).

fjosw:
This PR removes the filtering of `idls` in `derived_observable`.

Closes #138

### DIFF
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -57,7 +57,6 @@ class Obs:
     tau_exp_dict = {}
     N_sigma_global = 1.0
     N_sigma_dict = {}
-    filter_eps = 1e-10
 
     def __init__(self, samples, names, idl=None, **kwargs):
         """ Initialize Obs object.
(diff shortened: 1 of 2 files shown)


s-kuberski:
This looks good. Do you want to delete the function entirely or should we maybe put it in a separate function, e.g., in misc?

fjosw:
I think copying over this function to `misc` alone wouldn't add functionality, we would need a function that takes an `Obs` as an argument and returns a filtered `Obs`. I don't have a use case for such a function in mind but feel free to add it.

PR #140, opened 2023-01-06
Remove attribute is_merged

fjosw:
This PR removes the `Obs` attribute `is_merged` which is no longer needed after the removal of `_filter_zeroes`. On my machine, on python 3.8 this results in a ~1.15x speed up for the multiplication of two `Obs`.

### DIFF
--- a/pyerrors/input/json.py
+++ b/pyerrors/input/json.py
@@ -41,8 +41,6 @@ def create_json_string(ol, description='', indent=1):
             for r_name in ol[0].e_content[name]:
                 rd = {}
                 rd['name'] = r_name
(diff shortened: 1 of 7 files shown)


s-kuberski:
Given the speed up, we should probably get rid of this attribute... Therefore, this is fine with me, although I liked to have this extra piece of information in the past.

fjosw:
I agree, that the speed-up outweighs the missing info. I hope that this change does not break any workflows. The only case in which I see problems is trying to unpickle an old pyerrors 2.x pickle file with the new version but I think we made clear that this should be avoided anyways.

PR #144, opened 2023-01-14
better docstrings

jkuhl-uni:
Inspired by the automatically generated docstrings from my last pull request and some questions that I got from someone working with the fit functions, I started adding Return statements in .addition to the usual parameter statements.
So far I have worked on the input methods. Of course, this is only a part of the full job.
I intend to add more of these in the future, if there is interest in having this. In my opinion this could be quite practical for people who are new to the code and who work with IDEs that show doctrings as hints (e.g. VS Code).
In fact I noticed, that there are already statements like these in the docstrings for some methods, however I thought having them in the same standardized format as the parameters could be good.
Of course I'd be more than happy for additional input on this matter. Especially in the cases in which I have not written the input methods, it could be good to have another pair of eyes looking at this.

### DIFF
--- a/pyerrors/dirac.py
+++ b/pyerrors/dirac.py
@@ -26,6 +26,11 @@ def epsilon_tensor(i, j, k):
     """Rank-3 epsilon tensor
 
     Based on https://codegolf.stackexchange.com/a/160375
+
+    Returns
+    -------
+    elem : int
+        Element (i,j,k) of the epsilon tensor of rank 3
     """
     test_set = set((i, j, k))
     if not (test_set <= set((1, 2, 3)) or test_set <= set((0, 1, 2))):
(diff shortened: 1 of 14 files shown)


fjosw:
Hi Justus, thanks a lot, I really appreciate this! For the future I would suggest to gather these changes in fewer commits with meaningful commit messages (or squash them if that suits your workflow better).

jkuhl-uni:
Hi Fabian, 
sure, I'll carry on having this in mind.

Just to clarify: I also haven't worked on help-functions in the modules and class-functions like .gamma_method() or .reweight() as they are not a priority for this in my opinion. (the first ones are only for internal use, the second ones just change the object they are called from) Of course I'd be happy to come up with something similar for these functions too if this is of interest.

jkuhl-uni:
I guess this could be the first version of the docstrings. As mentioned above, there are still a lot of internal and class functions that have not been changed yet... Is this of interest?
Also, maybe someone else could have a look at the docstrings too, just in case there is any mistake (or something is badly described).
Otherwise this can obviously done later when this becomes apparent while working with it.

PR #146, opened 2023-01-18
Wrapper for errorbar matplotlib method

fjosw:
I added a wrapper around `plt.errorbar` or `Axes.errorbar` for use with `Obs` objects which should save a bit of time everytime one tries to plot and Obs valued list or array.

Usage:
```python
import matplotlib.pyplot as plt
import pyerrors as pe

...

pe.errorbar(x, y, marker="o")
plt.show()

or

fig = plt.figure()
ax1 = fig.add_subplot(111)
pe.errorbar(x, y, marker="o", axes=ax1)
plt.show()
```

### DIFF
--- a/pyerrors/misc.py
+++ b/pyerrors/misc.py
@@ -4,38 +4,35 @@ import matplotlib.pyplot as plt
 from .obs import Obs
 
 
-def errorbar(x, y, axes=plt, *args, **kwargs):
-    """pyerrors wrapper for the errorbars method fo matplotlib
+def errorbar(x, y, axes=plt, **kwargs):
+    """pyerrors wrapper for the errorbars method of matplotlib
 
     Parameters
     ----------
     x : list
-        A list of x-values. It can be a list of Obs objects or int/float.
+        A list of x-values which can be Obs.
     y : list
-        A list of y-values. It should be a list of Obs objects.
+        A list of y-values which can be Obs.
     axes : (matplotlib.pyplot.axes)
         The axes to plot on. default is plt.
     """
-    if not all(isinstance(o, Obs) for o in y):
-        raise Exception("All entries of 'y' have to be Obs")
+    val = {}
+    err = {}
+    for name, comp in zip(["x", "y"], [x, y]):
+        if all(isinstance(o, Obs) for o in comp):
+            if not all(hasattr(o, 'e_dvalue') for o in comp):
+                [o.gamma_method() for o in comp]
+            val[name] = [o.value for o in comp]
+            err[name] = [o.dvalue for o in comp]
+        else:
+            val[name] = comp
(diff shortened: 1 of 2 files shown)


s-kuberski:
Great idea! This is a very useful feature that will definitely save a lot of typing on my side. You could think of making the routine a bit more flexible and don't require all components to be either of type Obs or not. In the past, I ran into problems when trying to plot the content of a Corr that might have values that are `None`. 
One could use
```python
xval = [o.value if isinstance(o, Obs) else o for o in x]
xerr = [o.dvalue if isinstance(o, Obs) else None for o in x]
if all([o is None for o in xerr]):
    xerr = None

yval = [o.value if isinstance(o, Obs) else o for o in x]
yerr = [o.dvalue if isinstance(o, Obs) else None for o in x]
if all([o is None for o in yerr]):
    yerr = None
```
or even implement and utilise the short routines:
```python
def obsval(o):
    if isinstance(o, Obs):
        return o.value
    else:
        return o

def obsdval(o):
    if isinstance(o, Obs):
        return o.dvalue
    else:
        return None
```
that could be useful in other cases (I use the first one to be flexible in some cases).

I think a drop-in replacement of plt.errorbar that can cope with Obs but also works in all other cases would be even more valuable. In this case, the user could provide `xerr` and `yerr` and only in case there are not provided, they would be generated from the Obs in `x` and `y`.

As a last point, one could get the case to work where y is a Corr, i.e., cast the content to an array first. This is sort of done in `Corr.show()` but I guess it would not hurt to have it here, as well. I tend to be confused by having something like `[o[0] for o in c.content]` in my code. I guess this is also sort of implemented in `Corr.plottable`, so one could just use the routine here, if 'y' is a `Corr`.

fjosw:
Hi @s-kuberski I started looking into this but I think `None` entries in the x- and y-errors are incompatible
```python
plt.errorbar([2, None, 2], [3, None, 1], yerr=[0.1, None, 0.1])
TypeError: unsupported operand type(s) for *: 'int' and 'NoneType'
```

One could replace `None` entries with zero but that could be considered unexpected behavior.

Making the code work also with `xerr` and `yerr` as arguments sounds like a good idea, I will try to implement this.

s-kuberski:
You are right, `errorbar()` seems to be more restrictive than `plot()`. The latter can cope with `None` entries. Sorry for not having a look at this myself before suggesting it. 
I agree that setting the errors to zero could be considered undefined behavior. So a mixture of `Obs` and numbers is most likely not a good idea. A filtering of `None` entries in `y`, as done in `plottable` could be nice, but in the end, the user could also just do this manually, before passing the list to the routine.

s-kuberski:
This looks very nice!

PR #151, opened 2023-02-08
Impr/refactor find files

jkuhl-uni:
In this PR, I refactored the rad methods from the openQCD module. Simply put, I tried to make a more general standardized function to find the files to read.
I also added some tests to accommodate for the changes, as well as adding a test for the read_ms5_method.
I tried to let most things as they were, but I slightly changed the interface with respect to the "postfix" keyword. The interface should however still work in the old fashion as well


pyerrors/input/openQCD.py
```diff
@@ -154,7 +184,7 @@ def read_rwms(path, prefix, version='2.0', names=None, **kwargs):
                 nsrc.append(struct.unpack('i', t)[0])
             if version == '2.0':
                 if not struct.unpack('i', fp.read(4))[0] == 0:
-                    print('something is wrong!')
+                    print(' something is wrong!')
```
fjosw, line 187: Hi Justus, should this line raise an Exception or a Warning? I think the print statement is not very instructive to the user.

jkuhl-uni: Ah, this should probably be an exception. What happens here (to my understanding) is that there should be a 0 in the case of openQCD 2.0. Why this is, I am not sure (maybe @s-kuberski has more insight on this). However I agree, that there should probably be an error message like "Are you sure you are using the right version tag?" or something similar.

tests/openQCD_in_test.py
```diff
@@ -7,7 +7,7 @@
 def test_rwms():
     path = './tests//data/openqcd_test/'
     prefix = 'sfqcd'
-    postfix = '.rwms'
```
fjosw, line 10: Does this PR change the behavior of the input functions? Would using `.rwms` still work?

jkuhl-uni: Originally, it did, however after I realized this would be a bit unwieldy and it would be pretty easy to correct, I corrected it here: https://github.com/fjosw/pyerrors/pull/151/commits/fbf13493624023014c4d279bc391c9136ebe8bbf. Like this, everything works, ```rwms```, ```.rwms```, ```rwms.``` (whoever would use this) and ```.rwms.```

fjosw: Very good, I think the test should then cover a few of these cases.

jkuhl-uni: Sure, why not... however I would suggest to put this into the `test_find_files`, as that makes it simpler for later bug-fixing.

fjosw (approved):
Looks good, thanks!

PR #153, opened 2023-02-15
Impr/refactor find files

jkuhl-uni:
Due to some small issues by one user, if one only specifies very few characters as the prefix in the input methods, I built a better sorting algorithm for the files and replica names.
While working on this, I also found a minor bug, where the files of the replica themselves, but not the replica names would be sorted in one of the routines.
A small test for this is also submitted.
Please feel free to comment if this does not seem to work correctly or I forgot any edge-cases.
The ultimate fallback for this is still the same sorting used in earlier versions of the methods.


fjosw:
Hi @jkuhl-uni thanks for the fix. There seem to be conflicts between this PR and the develop branch. Can you please resolve them so that I can review the changes?

fjosw (approved):
Looks good, thanks!

PR #154, opened 2023-02-23
refactor: _standard_fit method made redundant.

fjosw:
I started refactoring the fit code and removed the `_standard_fit` function by merging its functionality with `_combined_fit`. All tests pass so it looks like nothing went wrong in the process but please have a close look @s-kuberski & @PiaLJP .

### DIFF
--- a/pyerrors/fits.py
+++ b/pyerrors/fits.py
@@ -516,6 +516,7 @@ def _combined_fit(x, y, func, silent=False, **kwargs):
     elif (type(x) == dict or type(y) == dict or type(func) == dict):
         raise TypeError("All arguments have to be dictionaries in order to perform a combined fit.")
     else:
+        x = np.asarray(x)
         xd = {"a": x}
         yd = {"a": y}
         funcd = {"a": func}
--- a/pyerrors/fits.py
+++ b/pyerrors/fits.py
@@ -168,13 +168,8 @@ def least_squares(x, y, func, priors=None, silent=False, **kwargs):
     '''
     if priors is not None:
         return _prior_fit(x, y, func, priors, silent=silent, **kwargs)
-
-    elif (type(x) == dict and type(y) == dict and type(func) == dict):
-        return _combined_fit(x, y, func, silent=silent, **kwargs)
-    elif (type(x) == dict or type(y) == dict or type(func) == dict):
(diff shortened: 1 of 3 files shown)


pyerrors/fits.py
```diff
@@ -916,11 +730,18 @@ def chisqfunc_compact(d):
...
         n_cov = np.min(np.vectorize(lambda x_all: x_all.N)(y_all))
         output.t2_p_value = 1 - scipy.stats.f.cdf((n_cov - output.dof) / (output.dof * (n_cov - 1)) * output.chisquare,
                                                   output.dof, n_cov - output.dof)
 
+    if kwargs.get('resplot') is True:
+        residual_plot(x, y, func, result)
+
+    if kwargs.get('qqplot') is True:
+        qqplot(x, y, func, result)
+
```
s-kuberski, line 744: This does not work if dictionaries are passed to `least_squares`. I guess `residual_plot` and `qqplot` have to be adapted to dictionaries, as well.

s-kuberski (changes requested):
Thank you for the changes. I spotted a bug that is not caught by the tests.

pyerrors/fits.py
fjosw: Thanks for spotting this. My changes should make the two routines work with combined fits.

s-kuberski:
Thanks! The two functions `residual_plot` and `qqplot` are invalid for correlated fits. This has been the case before and I don't say that we need to resolve this issue now. If, at some point, we want to show the transformed residuals that enter a correlated fit, we would have to supply `chol_inv`  and the complete `x` and `y` vectors, right?

fjosw:
I would say that `residual_plot` is not invalid, one just would not expect that the residuals are distributed ~ N(0, 1) if there are correlations between the data points. I think one could transform the residuals, but I am unsure if the transformation and the resulting visual representation are unambiguous. I would propose leaving it as it is for now.

s-kuberski:
I agree. We can keep this in mind for the future.

s-kuberski (approved):

PiaLJP:
Maybe it would be a good idea to add a warning then, along the lines of  'For correlated data points the residuals are not necessarily expected to be ~N(0,1)'? Apart from that I also think it looks good now, thanks.

PR #155, opened 2023-02-24
feat: added __slots__ to Corr class.

fjosw:
This PR adds the `__slots__` class attribute to the `Corr` class which should result in a few percent memory saving per `Corr` instance.


s-kuberski (approved):
Looks good!

PR #157, opened 2023-02-26
_calculate_drho for large time separations

### LINKED ISSUE #156: Gamma_method() is broken for Obs that are NaN
Commit #152 breaks the `gamma_method` for Obs that are NaN. 
Minimal example:
```python
o = pe.pseudo_Obs(1, .1, 'test')
o.gamma_method()
no = np.nan * o
no.gamma_method()
```
works fine in the old version but now throws an error:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-bd59fd2172b6> in <module>
      2 o.gamma_method()
      3 no = np.nan * o
----> 4 no.gamma_method()

~/phd/git/pyerrors_github/pyerrors/obs.py in gamma_method(self, **kwargs)
    323                     for n in range(1, w_max):
    324                         if g_w[n - 1] < 0 or n >= w_max - 1:
--> 325                             _compute_drho(gapsize * n)
    326                             n *= gapsize
    327                             self.e_tauint[e_name] = self.e_n_tauint[e_name][n] * (1 + (2 * n / gapsize + 1) / e_N) / (1 + 1 / e_N)  # Bias correction hep-lat/0306017 eq. (49)

~/phd/git/pyerrors_github/pyerrors/obs.py in _compute_drho(i)
    290 
    291             def _compute_drho(i):
--> 292                 tmp = self.e_rho[e_name][i + 1:w_max] + np.concatenate([self.e_rho[e_name][i - 1::-1], self.e_rho[e_name][1:w_max - 2 * i]]) - 2 * self.e_rho[e_name][i] * self.e_rho[e_name][1:w_max - i]
    293                 self.e_drho[e_name][i] = np.sqrt(np.sum(tmp ** 2) / e_N)
    294 

ValueError: operands could not be broadcast together with shapes (0,) (500,) 
```

This had not been the case before. However, If I revert the change, I can create a similar behavior when I add 
```python 
o.idl['test'] = [1, 5] + list(range(7, 2002, 2))
o.gamma_method()
no = np.NaN * o
no.gamma_method()
```
, i.e., when I use a gapped MC history with a hole, the following exception is thrown:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-3-4fdb987095af> in <module>
      8 o.gamma_method()
      9 no = np.NaN * o
---> 10 no.gamma_method()

~/phd/git/pyerrors_github/pyerrors/obs.py in gamma_method(self, **kwargs)
    323                     for n in range(1, w_max):
    324                         if n < w_max // 2 - 2:
--> 325                             _compute_drho(gapsize * n + gapsize)
    326                         if g_w[n - 1] < 0 or n >= w_max - 1:
    327                             n *= gapsize

~/phd/git/pyerrors_github/pyerrors/obs.py in _compute_drho(i)
    290 
    291             def _compute_drho(i):
--> 292                 tmp = self.e_rho[e_name][i + 1:w_max] + np.concatenate([self.e_rho[e_name][i - 1::-1], self.e_rho[e_name][1:w_max - 2 * i]]) - 2 * self.e_rho[e_name][i] * self.e_rho[e_name][1:w_max - i]
    293                 self.e_drho[e_name][i] = np.sqrt(np.sum(tmp ** 2) / e_N)
    294 

ValueError: operands could not be broadcast together with shapes (499,) (500,) 
```

fjosw:
This PR fixes `_compute_drho` for `i>w_max // 2` and closes #156. Irrespective of this we should maybe think about how the `gamma_method` should deal with observables that have values like `nan` and `inf`.

### DIFF
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -21,6 +21,6 @@ jobs:
       - name: flake8 Lint
         uses: py-actions/flake8@v2
         with:
-          ignore: "E501,E121,E123,E126,E226,E24,E704,W503,W504"
+          ignore: "E501,W503"
           exclude: "__init__.py, input/__init__.py"
           path: "pyerrors"
--- a/.github/workflows/flake8.yml
+++ b/.github/workflows/flake8.yml
@@ -21,6 +21,6 @@ jobs:
       - name: flake8 Lint
         uses: py-actions/flake8@v2
         with:
-          ignore: "E501"
+          ignore: "E501,E121,E123,E126,E226,E24,E704,W503,W504"
           exclude: "__init__.py, input/__init__.py"
           path: "pyerrors"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,6 +33,6 @@ pytest --cov=pyerrors --cov-report html
 ```
 The linter `flake8` is executed with the command
 ```
-flake8 --ignore=E501 --exclude=__init__.py pyerrors
+flake8 --extend-ignore=E501 --exclude=__init__.py pyerrors
 ```
 Please make sure that all tests are passed for a new pull requests.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -33,6 +33,6 @@ pytest --cov=pyerrors --cov-report html
 ```
 The linter `flake8` is executed with the command
 ```
-flake8 --extend-ignore=E501 --exclude=__init__.py pyerrors
+flake8 --ignore=E501,W503 --exclude=__init__.py pyerrors
 ```
 Please make sure that all tests are passed for a new pull requests.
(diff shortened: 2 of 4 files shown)


s-kuberski:
Great, thanks! This seems to resolve the first issue. 

In principle we could skip the whole procedure if the mean value is `nan` or `inf`.

The second issue in #156  is not yet resolved. I think the point is that we only want to go up to `w_max // gapsize` in the loop, because we use the indexing `[gapsize::gapsize]` here. This issue shows up for the `nan` obs, because the loop does not end before the last index is reached.
Would you like to add this in your pr?

fjosw:
Thanks for spotting this I think the last commit should cover this case. I also added your example to the tests.

PR #160, opened 2023-03-03
Fix/multi xdim fits

### LINKED ISSUE #159: Multi-dimensional fits
As anticipated, we have to polish the new fit routine: The recent commit #154 seems to lead to problems with fits where several independent observables are used. Minimal example:
```python
x1 = np.arange(1, 10)
x = np.array([[xi, xi] for xi in x1]).T
y = [pe.pseudo_Obs(i + 2 / i, .1*i, 't') for i in x[0]]
[o.gm() for o in y]
def fitf(a, x):
    return a[0] * x[0] + a[1] / x[1]

pe.fits.least_squares(x, y, fitf)
```
gives me
```python
---------------------------------------------------------------------------
Exception                                 Traceback (most recent call last)
<ipython-input-2-b5c7b1c30017> in <module>
      6     return a[0] * x[0] + a[1] / x[1]
      7 
----> 8 pe.fits.least_squares(x, y, fitf)

~/phd/git/pyerrors_github/pyerrors/fits.py in least_squares(x, y, func, priors, silent, **kwargs)
    170         return _prior_fit(x, y, func, priors, silent=silent, **kwargs)
    171     else:
--> 172         return _combined_fit(x, y, func, silent=silent, **kwargs)
    173 
    174 

~/phd/git/pyerrors_github/pyerrors/fits.py in _combined_fit(x, y, func, silent, **kwargs)
    556             raise TypeError('func (key=' + key + ') is not a function.')
    557         if len(xd[key]) != len(yd[key]):
--> 558             raise Exception('x and y input (key=' + key + ') do not have the same length')
    559         for i in range(100):
    560             try:

Exception: x and y input (key=) do not have the same length
```
In the old version, the comparison has been done using
```python 
if x.shape[-1] != len(y):
        raise Exception('x and y input have to have the same length')
```
so something along the lines of 
```python
if xd[key].shape[-1] != len(y[key]):
        raise Exception('x and y input (key = %s) have to have the same length' % (key))
```
should fix this.

fjosw:
This PR closes #159 and depends on #158

### DIFF
--- a/pyerrors/fits.py
+++ b/pyerrors/fits.py
@@ -554,7 +554,7 @@ def _combined_fit(x, y, func, silent=False, **kwargs):
     for key in key_ls:
         if not callable(funcd[key]):
             raise TypeError('func (key=' + key + ') is not a function.')
-        if len(xd[key]) != len(yd[key]):
+        if np.asarray(xd[key]).shape[-1] != len(yd[key]):
             raise Exception('x and y input (key=' + key + ') do not have the same length')
         for i in range(100):
             try:
--- a/tests/fits_test.py
+++ b/tests/fits_test.py
@@ -913,6 +913,17 @@ def test_combined_resplot_qqplot():
     plt.close('all')
 
 
+def test_x_multidim_fit():
+    x1 = np.arange(1, 10)
+    x = np.array([[xi, xi] for xi in x1]).T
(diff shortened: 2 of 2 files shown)


s-kuberski (approved):
Thanks for the changes (and the bug fix)! I did not have in mind that we need the uncorrelated chisquared to do the "pre fits" in the case of correlated fits such we cannot eliminate those completely. I still like the new version better than the old one, thanks for adapting it.

PR #161, opened 2023-03-06
Improved prior fit

### LINKED ISSUE #38: Constrained fit
Mathias made me take a closer look at constrained fits and I believe the constrained fits triggered by the keyword argument `const_par` do not work as expected. I attached a small example in which one can achieve the constraint also by subtracting a constant from the data. The errors for the unconstrained parameter differ by a relevant factor between the two methods. I have the assumption that the two should agree in the limit in which the error of the constraint is small compared to the error on the data.

```python
import numpy as np
import pyerrors as pe

dim = 10
x = np.arange(dim)
y = -0.06 * x + 5 + np.random.normal(0.0, 0.3, dim)
yerr = [0.3] * dim

oy = []
for i, item in enumerate(x):
    oy.append(pe.pseudo_Obs(y[i], yerr[i], 'test'))

def func(a, x):
    y = a[0] * x + a[1]
    return y

my_constant = pe.pseudo_Obs(5, 0.0001, 'test')

# Fit with constrained parameter
out = pe.least_squares(x, oy, func, const_par=[my_constant])

def alt_func(a, x):
    y = a[0] * x
    return y

alt_y = np.array(oy) - my_constant
[o.gamma_method() for o in alt_y]

# Fit with the constant subtracted from the data
alt_out = pe.least_squares(x, alt_y, alt_func)

print(out, alt_out)
print(out[0] == alt_out[0])
```

fjosw:
In this PR I combined `_combined_fit` and `_prior_fit` and eliminated duplicate code.
I also extended the prior fit functionality. Priors can now also be provided via a dictionary with integer keys. For example:
```python
prior_dict = {0: "2.0(7)",
              2: "1.12(2)"}
pe.least_squares(x, y, fitf, priors=prior_dict)
```
would only impose priors for parameters 0 and 2 and leave all other parameters unconstrained.

I think this PR closes #38 I added a test that demonstrates that constraining a constant fit parameter with a very precise prior is equivalent to subtracting the prior from the data.

Please have a close look at the changes and do some tests of your own if possible as this is quite a substantial change to the fitting functionality together with the changes in #154 and #158

### DIFF
--- a/tests/fits_test.py
+++ b/tests/fits_test.py
@@ -940,6 +940,67 @@ def test_x_multidim_fit():
     pe.fits.least_squares(x, y, fitf)
 
 
+def test_priors_dict_vs_list():
+    x = np.arange(1, 5)
+    y = [pe.pseudo_Obs(2 * i + 1.1 + np.random.normal(0.0, 0.01), .01, 't') for i in x]
+    [o.gm() for o in y];
+    def fitf(a, x):
+        return a[0] * x + a[1]
+    priors = [pe.cov_Obs(1.0, 0.0001 ** 2, "p0"), pe.cov_Obs(1.1, 0.8 ** 2, "p1")]
+    pr1 = pe.fits.least_squares(x, y, fitf, priors=priors)
+    prd = {0: priors[0],
+           1: priors[1]}
+
+    pr2 = pe.fits.least_squares(x, y, fitf, priors=prd)
+
+    prd = {1: priors[1],
+           0: priors[0]}
+
+    pr3 = pe.fits.least_squares(x, y, fitf, priors=prd)
+    assert (pr1[0] - pr2[0]).is_zero(1e-6)
+    assert (pr1[1] - pr2[1]).is_zero(1e-6)
+    assert (pr1[0] - pr3[0]).is_zero(1e-6)
+    assert (pr1[1] - pr3[1]).is_zero(1e-6)
+
+
+def test_not_all_priors_set():
+    x = np.arange(1, 5)
+    y = [pe.pseudo_Obs(2 * i + 1.1 + np.random.normal(0.0, 0.01), .01, 't') for i in x]
+    [o.gm() for o in y];
+    def fitf(a, x):
+        return a[0] * x + a[1] + a[2] * x ** 2
+    priors = [pe.cov_Obs(2.0, 0.1 ** 2, "p0"), pe.cov_Obs(2, 0.8 ** 2, "p2")]
+    prd = {0: priors[0],
+           2: priors[1]}
+
+    pr1 = pe.fits.least_squares(x, y, fitf, priors=prd)
+
+    prd = {2: priors[1],
+           0: priors[0]}
+
+    pr2 = pe.fits.least_squares(x, y, fitf, priors=prd)
+    assert (pr1[0] - pr2[0]).is_zero(1e-6)
+    assert (pr1[1] - pr2[1]).is_zero(1e-6)
+    assert (pr1[2] - pr2[2]).is_zero(1e-6)
+
+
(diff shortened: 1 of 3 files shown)


fjosw:
I just ran a few benchmarks vs the previous version and it looks like the new implementation is about 15-25% slower for a standard two-parameter fit without priors. I suspect that this is due to the additional `concatenate` in the chisquare. The question is if we are fine with this or want to add an additional if case to get back the original speed.

tests/fits_test.py
```diff
@@ -924,6 +940,107 @@ def fitf(a, x):
...
+
+        oy = []
+        for i, item in enumerate(x):
+            oy.append(pe.pseudo_Obs(y[i], yerr[i], 'test'))
+
+        def func(a, x):
+            y = a[0] * x + a[1]
+            return y
+
+        my_constant = pe.pseudo_Obs(6.5, 0.00000001, "test")
```
s-kuberski, line 1022: This makes the for loop trivial.

fjosw: Thanks for spotting this!

s-kuberski:
Thanks! I am in the process of checking your changes and stumbled upon another issue: When x is a dict and its entries are lists instead of arrays, the routine `resplot` does not work. I think this is because what is done in https://github.com/fjosw/pyerrors/blob/4098994ea67a1a463b9d58728ea91cbad9de2f02/pyerrors/fits.py#L180 is not performed for the entries of the dictionary. If you write
``` python
xd = {k: np.asarray(x[k]) for k in x}
```
instead of https://github.com/fjosw/pyerrors/blob/4098994ea67a1a463b9d58728ea91cbad9de2f02/pyerrors/fits.py#L173
this should be fixed and in turn, we can probably remove some calls to `asarray` that float around in the routine.

My minimal (not) working example is 
```  python
xd = {
    'a': [1, 2, 3],
    'b': [1, 2, 3],
}
yd = {
    'a': [pe.pseudo_Obs(i, .1*i, 't') for i in range(1, 4)],
    'b': [pe.pseudo_Obs(2*i**2, .1*i**2, 't') for i in range(1, 4)]
}
[[o.gamma_method() for o in y] for y in yd.values()]
fd = {
    'a': lambda a, x: a[0] + a[1] * x,
    'b': lambda a, x: a[0] + a[2] * x**2,
}

fitp = pe.fits.least_squares(xd, yd, fd, resplot=True)
```
which is resolved, if I convert the entries of `xd` to arrays.

s-kuberski:
> I just ran a few benchmarks vs the previous version and it looks like the new implementation is about 15-25% slower for a standard two-parameter fit without priors. I suspect that this is due to the additional `concatenate` in the chisquare. The question is if we are fine with this or want to add an additional if case to get back the original speed.

Probably the fact that you need to allocate memory for a second time is causing this. I did not yet come up with a better idea to easily fix this (apart from your `if`). It definitely makes sense to optimize the evaluation of the chisquare function as much as possible.

s-kuberski:
> Thanks for spotting another bug, I adopted your suggestion and added your test

Very good, thanks! I did not think of the fact that this problem now appears if I call resplot explicitly, after the fit has been done, 
``` python
for key in xd:
    pe.fits.residual_plot(xd[key], yd[key], fd[key], fitp.fit_parameters, title=key)
```
If we want to make usage as easy as possible, we should probably introduce another `np.asarray()` in https://github.com/fjosw/pyerrors/blob/4098994ea67a1a463b9d58728ea91cbad9de2f02/pyerrors/fits.py#L742

fjosw:
I hope the last commit fixes this issue.

s-kuberski (approved):

fjosw:
Thanks, I will merge the code but wait a bit with the next release such that we can test a bit more on develop.

PR #165, opened 2023-03-16
fixed appended mode reading

jkuhl-uni:
In #163 there was still somethign missing, now the appended mode data is also sorted according to `sort_names`


fjosw (approved):
Great, thanks!

PR #168, opened 2023-03-24
extended extract_t0 by postfix argument, default stays .ms

jkuhl-uni:
Hi,
contrary to the currently implemented solution, I need the postfix used for the extraction of t0 to be .ms3.
I therefore did a small fix, in which you can now optionally also hand over a postfix to this method, similar to what is already implemented in read_rwms().
The default is still .ms, as this would break backwards compatibility otherwise.
PS: Sorry for the typo in the branch name :)


fjosw (approved):
Looks good, thanks!

PR #170, opened 2023-04-20
Feat/str format

fjosw:
This PR adds a `__format__` method to the `Obs` class and extends the functionality of `_format_uncertainty`.
For example the following code
```python
o0 = pe.pseudo_Obs(-0.3, 0.421, "test")
o1 = pe.pseudo_Obs(0.21, 0.1452879, "test")
for i, o in enumerate([o0, o1]):
    print(f"Obs{i} {o:+3}")
```
now outputs three significant digits and adds a + sign to non-negative observables:
```bash
Obs0 -0.300(421)
Obs1 +0.210(145)
```

### DIFF
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -707,8 +707,9 @@ class Obs:
         return _format_uncertainty(self.value, self._dvalue)
 
     def __format__(self, format_type):
-        my_str = self.__str__()
-        if format_type in ["+", "-"]:
+        my_str = _format_uncertainty(self.value, self._dvalue,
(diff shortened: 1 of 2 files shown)


fjosw:
Thanks for the feedback. I now implemented `"+"`, `"-"` and `" "` as in the documentation. For now I will not look into scientific notation or alignment. If needed alignment could be achieved with nested f-strings, for example:
```python
print(f"{f'{o1:+3}':<20}text")
```
```bash
+0.3480(123)        text
```

PR #171, opened 2023-04-21
t0 extractions for new Hadrons module

fjosw:
I refactored the fit to extract t0 into `fit_t0` and added a new function that can extract t0 from a file generated by the new Hadrons module FlowObservables (see https://github.com/aportelli/Hadrons/pull/108).

### DIFF
--- a/pyerrors/input/hadrons.py
+++ b/pyerrors/input/hadrons.py
@@ -121,6 +121,67 @@ def read_meson_hd5(path, filestem, ens_id, meson='meson_0', idl=None, gammas=Non
     return corr
 
 
+def _extract_real_arrays(path, files, tree, keys):
+    corr_data = {}
+    for key in keys:
+        corr_data[key] = []
+    for hd5_file in files:
+        h5file = h5py.File(path + '/' + hd5_file, "r")
+        for key in keys:
+            if not tree + '/' + key in h5file:
+                raise Exception("Entry '" + key + "' not contained in the files.")
+            raw_data = h5file[tree + '/' + key + '/data']
+            real_data = raw_data[:].astype(np.double)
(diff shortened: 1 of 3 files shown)


s-kuberski:
Hi. Looks good! I always thought about refactoring the openQCD routine by calling another routine that does the I/O and extracts the Yang-Mills action density Y. This routine could then also be used by a user of the package. However, the user would probably extract Y at flow time t_0, such that one would have a somewhat circular dependence.
If you consider this useful also for your purposes (Y is a good observable to look at autocorrelation), you could add this feature for the hadrons and openQCD routines. Otherwise, I am totally fine with your current changes.

jkuhl-uni:
> Hi. Looks good! I always thought about refactoring the openQCD routine by calling another routine that does the I/O and extracts the Yang-Mills action density Y. This routine could then also be used by a user of the package. However, the user would probably extract Y at flow time t_0, such that one would have a somewhat circular dependence. If you consider this useful also for your purposes (Y is a good observable to look at autocorrelation), you could add this feature for the hadrons and openQCD routines. Otherwise, I am totally fine with your current changes.

Hi, first of all @fjosw thank you for the changes. That is a good idea.
Concerning @s-kuberski's comment: When I had a look at the openQCD module last time, I already thought about something similar, but more general... couldn't we make a general purpose functio for reading openQCD formats, from which we can then query the value of interest from a list sorted by the configuration number?
Maybe we should discuss this in s different issue, but do you think somthing along those lines is possible?

fjosw:
Thanks for the feedback @s-kuberski and @jkuhl-uni. In this PR I only intended to add the t0 extraction to the hadrons module. I don't have any plans to restructure the openQCD module for now but feel free to work on it!

PR #173, opened 2023-05-03
fix: String conversion of Obs can now handle a dvalue that is NaN or inf

s-kuberski:
Calling 
```python
pe.obs._format_uncertainty(1, np.NaN)
```
led to an error because the case where dvalue is `nan` or `inf` has not been handled. This pr fixes this issue.


s-kuberski:
This assumes that the value is also `nan` (which should be the case, I guess) and that `nan` is printed. One could also explicitly handle this case by either checking if the value is `nan` or by explicitly printing the `nan` in brackets.

PR #176, opened 2023-05-10
Feat/flow

s-kuberski:
I added the extraction of w_0/a from the openQCD gradient flow files. To do this efficiently, I introduced a new internal routine `_extract_flowed_energy_density` that is now used by `extract_t0` and `extract_w0`.

The flowed Yang-Mills action density at flow time `t0` can be computed by `Yt = 0.3/t0**2`.

### DIFF
--- a/pyerrors/input/misc.py
+++ b/pyerrors/input/misc.py
@@ -2,6 +2,7 @@ import os
 import fnmatch
 import re
 import struct
+import warnings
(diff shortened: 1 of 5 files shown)


fjosw (changes requested):
Thanks, this looks good! Can you maybe add the missing doc string to `fit_t0`? Looks like a missed doing that.

fjosw (approved):
Thanks!

PR #177, opened 2023-05-16
Fix/sort names

jkuhl-uni:
Hi,
I found two small bugs in sort_names:
One is of cosmetic nature, the other one happened if only id was given as a sort pattern.
In that case, the method would subsequently use the fallback method.
While this did not harm the entire thing, it made some unneccessary work, so I corrected that and added tests for all cases.


fjosw (approved):
Looks good, thanks for taking care of this!

PR #184, opened 2023-05-22
taking care of cols with only None values

jkuhl-uni:
Hi,
I realised that #174 breaks compatibility with columns of a database that only have None values, this PR fixes that.


fjosw (approved):
Thanks!

PR #185, opened 2023-05-26
Feat/files idl xsf

jkuhl-uni:
Hi,
I made small changes to the xsf read method. Like this, the user can specify a range of idls that are to be read from each replikum file. This could be useful e.g. for crosschecks with other methods.
If the "expected" (meaning user-given) idls are not found in a file, a warning is thrown.
I also added another test for this case.


fjosw (approved):
Thanks!

PR #188, opened 2023-05-30
Adjusted maximal value for rho in test_gamma_method_irregular.

fjosw:
PR #172 introduced a warning in the tests as `rho` can become larger than one which results in the computation of the square root of a negative number. I adjusted the maximal value for `rho` to 1.


s-kuberski:
Looks good, I did not think about this.

s-kuberski (approved):

PR #189, opened 2023-05-30
Bug fix for edge case in _compute_drho

s-kuberski:
I finally got the indexing in the `_compute_drho` routine correct. My last fix did not solve all the edge cases that one can find and in one case, I ran into an error because the code tried to add arrays of non-equal size. I am fairly sure that the counting is now correct.

### DIFF
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -278,7 +278,7 @@ class Obs:
 
             def _compute_drho(i):
                 tmp = (self.e_rho[e_name][i + 1:w_max]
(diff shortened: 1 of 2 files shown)


fjosw (approved):
Looks good, thanks!

s-kuberski:
I tried to check this along the lines of
```python
w_max = 49
r = range(100000)
for i in range(1, w_max):
    lll = [len(r[i + 1:w_max]),
        len(r[i - 1:None if i - w_max // 2 <= 0 else 2 * i - (2 * w_max)//2 : -1]) + len(r[1:max(1, w_max - 2 * i)]),
        len(r[1:w_max - i])]
    assert(np.all([o - lll[0] == 0 for o in lll]))
```
which did throw an error with the old counting.

PR #192, opened 2023-06-01
fix nan to None test

### LINKED ISSUE #187: Warning in pandas tests
The recent changes in #174 and #184 seem to have introduced the following warning when calling the tests:
```
tests/pandas_test.py::test_nan_df_export_import
  /home/fjosw/Codes/pyerrors/pyerrors/input/pandas.py:85: UserWarning: nan value in column int will be replaced by None
    warnings.warn("nan value in column " + column + " will be replaced by None", UserWarning)
```
Can you please have a look at this @jkuhl-uni ?

jkuhl-uni:
Hi, this PR fixes the issue raised in #187.
In fact I just didn't read the pytest.warn docs not closely enough.
This alters the test such that this should not happen anymore.

### DIFF
--- a/tests/pandas_test.py
+++ b/tests/pandas_test.py
@@ -46,10 +46,9 @@ def test_nan_df_export_import(tmp_path):
     my_df.loc[1, "int"] = np.nan
 
     for gz in [True, False]:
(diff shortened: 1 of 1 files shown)


fjosw (approved):
Thanks! I will change the ci to fail on warnings so that we spot this type of behavior easier in the future.

PR #194, opened 2023-06-02
Fix another edge case in  _compute_drho

fjosw:
Some of my analysis scripts failed after the recent changes in #172. I added a minimal failing example and fixed the problem with another small modification in `_compute_drho`. All tests seem to be passing, please have a close look.


s-kuberski:
Ah, yes. We're getting there. In the test code that I produced in https://github.com/fjosw/pyerrors/pull/189#issuecomment-1568534437, the case that you have fixed would have resulted in an inconsistency for `w_max=48` and `i=24`.  You change resolves this.

s-kuberski (approved):

PR #197, opened 2023-07-07
new variant of second derivative, bigger stencil

jkuhl-uni:
Hello there,
I need another derivative for one of my projects, so I implemented it.
Of course, we can change the name, I just thought it would be practical to directly implement this is the `Corr` class.
If this is a use-case that is too niche I understand and you can discard this pull request.


fjosw (changes requested):
Looks good to me, I would be happy to include the additional derivative. Can you please fix the linting error and add an explanation of the new type of derivative to the docstring?

fjosw:
I think for the math to work you need to use a raw string (`r"""` or `r'''`).

jkuhl-uni:
I think you are right, I oversaw this

PR #198, opened 2023-07-11
Bootstrap export/import

fjosw:
I added the functionality to bootstrap resample an `Obs` with given random numbers or via random numbers generated from the ensemble name.

If the number of bootstrap samples is larger than the number of Monte Carlo samples the bootstrap samples can then be reimported into an `Obs`.

### DIFF
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -3,6 +3,7 @@ import hashlib
 import pickle
 import numpy as np
 import autograd.numpy as anp  # Thinly-wrapped numpy
+import scipy
 from autograd import jacobian
 import matplotlib.pyplot as plt
 from scipy.stats import skew, skewtest, kurtosis, kurtosistest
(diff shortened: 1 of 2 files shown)


pyerrors/obs.py
```diff
@@ -684,6 +685,49 @@ def export_jackknife(self):
...
+
+        Returns
+        -------
+        numpy.ndarray
+            Returns a numpy array of length N + 1 where N is the number of samples
+            for the given ensemble and replicum. The zeroth entry of the array contains
+            the mean value of the Obs, entries 1 to N contain the N jackknife samples
+            derived from the Obs. The current implementation only works for observables
+            defined on exactly one ensemble and replicum. The derived jackknife samples
+            should agree with samples from a full jackknife analysis up to O(1/N).
```
s-kuberski, line 709: I think it should be
jackknife --> bootstrap
right?

pyerrors/obs.py
```diff
@@ -1550,6 +1594,41 @@ def import_jackknife(jacks, name, idl=None):
     return new_obs
 
 
+def import_bootstrap(boots, name, shape, random_numbers=None):
```
s-kuberski, line 1597: In principle, one could allow to pass an `idl` here that is used when the Obs is constructed.

s-kuberski:
Thanks for adding this. I did only make very minor comments. 

Is it clear that you can unambiguously reconstruct the deltas when the number of Bootstrap samples is larger than the number of configs? In principle it could happen that one config is not drawn for all of the bins, right? I could also imagine other (rare) combinations of events that could prevent us from reconstructing the deltas (since we have an underdetermined system of equations) but maybe I am missing something and there is a solid proof...

While you are at it: You could add the `import_jackknife` routine that goes with the `export_jackknife` which is already present in the code, if you have time.

fjosw:
I had a look at how likely it is that the bootstrap projection matrix is rank deficient if the number of bootstrap samples is greater or equal to the number of Monte Carlo samples and my result is extremely unlikely. For a projection matrix of size 16^2 this happened once in 100_000 random experiments, for 25^2 I didn't see a single case in 1_000_000 random experiments. From this, I would expect that rank deficiency basically never happens for realistic numbers >=50.

s-kuberski:
Hi, 
thanks for the changes. Two very minor comments concerning the doc strings:
- There are still two appearances of `jackknife` in `export_bootstrap`
- `length` is not really defined in the docstring of `import_bootstrap`. While this could be self-explanatory, I think it could be useful to be explicit here.
- I somehow was not sure if `random numbers` is the best term to refer to the randomized configuration indices that are needed to construct the samples, but this should be from the context.

Thanks also for explicitly checking the reconstruction. I think this should be sufficient for all use cases.

PR #199, opened 2023-07-13
Matmul overloaded for correlator class.

fjosw:
I overloaded the matmul class which should now allow for the multiplication of matrix valued `Corr` objects with matrices. I will write a few tests and look into edge cases but this solution seems to work.

### DIFF
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -257,7 +257,7 @@ class Corr:
     def trace(self):
         """Calculates the per-timeslice trace of a correlator matrix."""
         if self.N == 1:
-            raise TypeError("Only works for correlator matrices.")
+            raise ValueError("Only works for correlator matrices.")
         newcontent = []
         for t in range(self.T):
             if _check_for_none(self, self.content[t]):
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -255,6 +255,7 @@ class Corr:
         return True
 
     def trace(self):
+        """Calculates the per-timeslice trace of a correlator matrix."""
         if self.N == 1:
             raise TypeError("Only works for correlator matrices.")
         newcontent = []
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -417,7 +417,7 @@ class Corr:
         dt : int
             number of timeslices
         """
-        return Corr(list(np.roll(np.array(self.content, dtype=object), dt)))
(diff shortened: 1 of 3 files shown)


fjosw:
I know extended the `matmul` method to also work on two correlator objects.

Additionally, I added a `trace` method for matrix correlators.

fjosw:
Another bug in `Corr.roll` for matrix correlators fixed.

pyerrors/correlators.py
```diff
@@ -254,6 +254,18 @@ def is_matrix_symmetric(self):
                         return False
         return True
 
+    def trace(self):
+        """Calculates the per-timeslice trace of a correlator matrix."""
+        if self.N == 1:
+            raise TypeError("Only works for correlator matrices.")
```
s-kuberski, line 260: Strictly speaking, it is not necessary to catch this, since the trace of a 1x1 matrix should be well defined. But of course it does not make much sense.

pyerrors/correlators.py
```diff
@@ -405,7 +417,7 @@ def roll(self, dt):
         dt : int
             number of timeslices
         """
-        return Corr(list(np.roll(np.array(self.content, dtype=object), dt)))
+        return Corr(list(np.roll(np.array(self.content, dtype=object), dt, axis=0)))
```
s-kuberski, line 420: Very nice!

s-kuberski:
Hi,

I did not (yet?) check the code by running it, but so-far everything looks fine. Thanks for taking care of this.

While you are at it: Currently, it is not possible to initialize a `Corr` object from a one-dimensional array:
```python
li = [pe.cov_Obs(o, 0, '') for o in [1, 2, 3, 4]]
ar = np.array(li)
pe.Corr(li)
pe.Corr(ar)
```
```
----> 4 pe.Corr(ar)

     46             # This only works, if the array fulfills the conditions below
---> 47             if not len(data_input.shape) == 2 and data_input.shape[0] == data_input.shape[1]:
     48                 raise Exception("Incompatible array shape")
     49             if not all([isinstance(item, Corr) for item in data_input.flatten()]):

IndexError: tuple index out of range
```
would you mind looking into this?

pyerrors/correlators.py
fjosw: I just checked and numpy does not like the trace of an array with a single entry:
```python
np.trace(np.array([1.1]))
> ValueError: diag requires an array of at least two dimensions
```

PR #200, opened 2023-07-13
CObs format added and complex Corr print improved.

fjosw:
I worked a bit on the string conversion of complex observables and correlators.

- `CObs` now have a `__format__` method which allows for f-string formatting like this: 
```python
print(f"{my_cobs:+3}")
> (+0.018(465)+0.443(393)j)
```
- I use this to also improve the print output of complex correlators which now produces the following:
```python
Corr T=4 N=1
x0/a	Corr(x0/a)
------------------
0	(+0.02(46)+0.44(39)j)
1	(+0.00(59)+0.43(37)j)
2	(+0.01(58)+0.64(47)j)
3	(-0.03(62)+0.45(51)j)
```
 Please have a look @JanNeuendorf

PS: There is still no proper print output for matrix correlators.

### DIFF
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -998,8 +998,6 @@ class Corr:
             content_string += "Description: " + self.tag + "\n"
         if self.N != 1:
             return content_string
-        if isinstance(self[0], CObs):
-            return content_string
 
         if print_range[1]:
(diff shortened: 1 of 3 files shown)


s-kuberski (approved):
Looks good!

PR #201, opened 2023-07-14
feat: Added numerical integration of generic functions

s-kuberski:
I have added a routine that allows to numerically integrate a one-dimensional function using scipy.integrate.quad. The parameters of the function, as well as the integration bounds, may be Obs.

@fjosw: I was not completely sure about the magic that is applied when constructing the `derived_observable` maybe you can check it, the current version seems to do the correct thing.

I think the tests should check the relevant cases as one can very nicely compare the numeric with the analytic integration. I hope that I did not overlook any edge case.


fjosw:
Looks good! I will do some explicit testing later.

Concerning the interface I would maybe propose to have a separate submodule `integrate` and then to rename the new function to `quad` to mimic the scipy interface. This would also make room for potential future implementations of other integration methods.

s-kuberski:
Thanks for looking into it. I refactored the code and added the new module.

fjosw (approved):
I played around a bit with the new function and everything looks good.

I made a small correction to the docstring and added two more trivial test cases.

s-kuberski:
There is one case, where this routine fails, namely when no Obs is involved, at all. In this case, `derived_observable` throws an exception.
One could circumvent this by adding 
```python
if len(derivint) == 0:
    return integration_result
```
before constructing the derived observable, but maybe you have something smarter in mind?

fjosw:
> There is one case, where this routine fails, namely when no Obs is involved, at all. In this case, `derived_observable` throws an exception. One could circumvent this by adding
> 
> ```python
> if len(derivint) == 0:
>     return integration_result
> ```
> 
> before constructing the derived observable, ~~but maybe you have something smarter in mind~~?

That sounds like a good solution, do you want to quickly push it?

PR #202, opened 2023-07-14
Explicit Exception for combined fit edge case

fjosw:
This PR addresses an issue in combined fits that @PiaLJP had seen in the past and @matteo-dc ran into this week:
When doing a combined fit and one of the fit functions is just a constant then the shape casting fails and one gets an error. The solution we found in the past is to add `x * 0` to the constant in the fit function to obtain the correct shape.

This PR does not fix the issue as I wasn't able to come up with a good solution, yet. Instead, I added an explicit exception that informs the user and suggests our workaround. For the future, it would of course be desirable to correct the shape casting for this edge case.

### DIFF
--- a/pyerrors/fits.py
+++ b/pyerrors/fits.py
@@ -230,6 +230,12 @@ def least_squares(x, y, func, priors=None, silent=False, **kwargs):
         n_parms_ls.append(n_loc)
 
     n_parms = max(n_parms_ls)
+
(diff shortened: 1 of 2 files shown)


s-kuberski:
Interesting problem. I'll try to think of a solution, but in the meanwhile, your addition to the Exception seems to be quite helpful.

PR #205, opened 2023-07-18
General hadrons hdf5 reader added

fjosw:
I added a more general hadrons hdf5 reader which can identify entries based on a dictionary of attributes. The function can also extract either the real or imaginary part or a complex correlator. `read_meson_hd5` now calls the new function preserving the old functionality. Compared to `read_meson_hd5` the new function has a cleaner interface with fewer arguments.

### DIFF
--- a/pyerrors/input/hadrons.py
+++ b/pyerrors/input/hadrons.py
@@ -54,6 +54,81 @@ def _get_files(path, filestem, idl):
     return filtered_files, idx
 
 
+def read_hd5(filestem, ens_id, group, attrs=None, idl=None):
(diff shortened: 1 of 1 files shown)


jkuhl-uni:
Yes, this looks very good. I understand it correctly that one would be able to read multiple correcators at once from the file  with the `attrs` dict, right?

jkuhl-uni:
Ah okay, yes... a second look at ll. 170 explained that...
Yes, I agree, it would not be very useful, except for one case, except for one weird case, which would be that the latency for reading the data from disc would be enormous. I thought about implenting this case for the SFCF read method.
We can maybe talk about that in an other discussion.

PR #206, opened 2023-07-18
__eq__ method for Corr class

fjosw:
I added a `__eq__` method for the `Corr` class for simpler checks like `corr == 0`.

### DIFF
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -1056,6 +1056,9 @@ class Corr:
 
     __array_priority__ = 10000
 
+    def __eq__(self, y):
+        return np.all([np.all(el == 0) or el is None for el in self - y])
+
     def __add__(self, y):
         if isinstance(y, Corr):
             if ((self.N != y.N) or (self.T != y.T)):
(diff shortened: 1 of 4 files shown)


fjosw:
My initial idea for implementing `__eq__` was to check whether the difference is zero on every timeslice leaning on the comparison functionality of the `Obs` class. Based on this comparisons are possible for every object that can be subtracted from a correlator.

I then noticed a few subtleties with `None` entries as these are ignored by the mathematical operations. For this reason, I wrote a special case for the comparison of two correlators where the `None` entries are explicitly checked. All other operations still fall back to the difference so the correlator can be compared with `int`, `float`, `Obs`, an array of `Obs` etc. although you are right, that the last case rarely evaluates to true.

fjosw:
I agree with you that this behavior can be confusing. I now rewrote the comparison to return a boolean array that has the same shape as the correlator. To get a single bool one could now use
```python
(corr1 == 0).all()
(corr1 == corr2).all()
```

The new behavior is closer to `np.ndarray` which I prefer as we set up the `Corr` class to be an array container with a dedicated 0th dimension. What do you think?

PR #207, opened 2023-07-18
Vectorized gamma_method

fjosw:
I added a vectorized `gamma_method` which should simplify applying the `gamma_method` to lists or arrays of `Obs`. I often used patterns like
```python
[o.gm() for o in obs_list]
```
This can now be replaced with
```python
pe.gm(obs_list)
```
Additionally, this also works for arrays with arbitrary dimensions.


s-kuberski:
Very nice, this will be useful! My only question would be if an alias `gamma_method` is needed, but I guess the newly introduced `gm` should be fine for all purposes.

PR #209, opened 2023-07-21
Added support for addition and multiplication of complex numbers to Corr objects

fjosw:
I added the missing overloading of the addition of `Obs` and `complex` and the addition and multiplication of `Corr` and `complex` types.


s-kuberski (approved):
Very good, thanks!

PR #211, opened 2023-10-24
Fixed bugs for combined fits with multiple independent variables

s-kuberski:
This fixes two bugs that were present in `least_squares` when doing a combined fit with multiple independent variables. The reason for the bug was that transposing the concatenation of the x values did not result in the expected shape. 

This fixes:
- The computation of the number of degrees of freedom works now correctly.
- The computation of the expected chi^2 does not throw an error anymore.

I furthermore discovered that the expected chi^2 with priors is not correctly evaluated. Therefore I have written the tests such that they don't detect this. I think I know what to do, but let's fix this bug first.


fjosw (approved):
Looks good, thanks for spotting this!

PR #212, opened 2023-10-27
[ci] Added python 3.12 to pytest and examples workflows.

fjosw:


fjosw:
python 3.12 ci fails because of
```
DeprecationWarning: datetime.datetime.utcfromtimestamp() is deprecated and scheduled for removal in a future version. Use timezone-aware objects to represent datetimes in UTC: datetime.datetime.fromtimestamp(timestamp, datetime.UTC).
```
Will fix that in a separate PR.

PR #221, opened 2023-12-08
Special function module

fjosw:
Here is a first draft for a `special` module which adds support for automatic differentation for the Bessel functions implemented in scipy. There are a few more functions which are already supported in `autograd` but I did not explicilty import them yet.

This module could also host functions which are not implemented in scipy, like for example the Lüscher Zeta functions.

### DIFF
--- a/pyerrors/__init__.py
+++ b/pyerrors/__init__.py
@@ -487,5 +487,6 @@ from . import linalg
 from . import mpm
 from . import roots
 from . import integrate
+from . import special
 
 from .version import __version__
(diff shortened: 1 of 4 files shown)


s-kuberski:
Hi,
thanks for starting this. I think it's a very good idea to include this in `pyerrors`. Just to have this as complete as possible, we could import everything that is already implemented in `autograd`. Also, we could add (somewhere) in the documentation that new functions can easily be added, when the derivatives are analytically known.

In view of lattice QCD workflows, having a good implementation of the Lüscher Zeta function with automatic differentiation would be a quite significant improvement of the package!

Issue #222, opened 2024-01-10
Read specific interval with read_ms5_xsf()


j-bode:
Hi,
read_ms5_xsf() in pyerrors/input/openQCD.py reads all configs/samples from the output file. It would be nice to define start and stop values in order to read just a subinterval, e.g. 10-20 instead of all lets say 1-500.

jkuhl-uni:
I'll close this issue, as the requested functionality was added in v2.9.0 and extended functionality was added in #224

PR #224, opened 2024-01-11
Throw error when trying to only find idls that are not present

jkuhl-uni:
As addressed in #222, this PR introduces an error in case none of the desired idls can be found in the file of the respective replica.


fjosw (approved):
Thanks, looks good. Closes #222

Issue #225, opened 2024-01-17
Files keyword for multiple reps in read_sfcf


jkuhl-uni:
Hi there,
just wanted to make you aware (and maybe have a small discussion) about an inconsistency in the `input.sfcf.read_sfcf_multi()` and therefore also in the `input.sfcf.read_sfcfi()` method. There exists a keyword `files`, which takes a list of files to be read. I noticed, that this only works for one replica and not if you have multiple ones.
I can take care of this, and it seems not that hard to me.  I'll make a PR when I am done.  Before that, I just wanted to check in for one thing:
This functionality can be implemented either as a list of lists, or, similar to `idl` as a dict of lists.
There are arguments for both implementations in my opinion. On one hand, this kind of functionality is implemented in this function via lists of lists. On the other hand, I think it is saver to implement it via a `dict`, as one can recheck, if the files are written in the right replica.
It is possible to implement both and a third option which covers the current (false-ish) implementation, such that we do not break any backwards compatibility.

s-kuberski:
Hi,
I don't really understand your comment concerning 'idl'. Do you mean the way this attribute is stored in an `Obs`? I think a list is fine and it is also in line with what is used for the openQCD input routines (where, of course, only one file per replicum is used).

jkuhl-uni:
I think we are talking about the same, yes. My alternative to lists would have been to use the same structure as used in `Obs.idl`. However, I agree that this change would change the user API by quite a bit and we can discard the idea.

PR #226, opened 2024-01-17
Fix keyword files with list of lists

jkuhl-uni:
Hi,
this is the first version of an implementation to solve #225. This is using a list of lists.
It can easily be adapted to use dicts.

### DIFF
--- a/pyerrors/input/sfcf.py
+++ b/pyerrors/input/sfcf.py
@@ -121,7 +121,7 @@ def read_sfcf_multi(path, prefix, name_list, quarks_list=['.*'], corr_type_list=
         String that separates the ensemble identifier from the configuration number (default 'n').
     replica: list
         list of replica to be read, default is all
-    files: list
+    files: list[list[int]]
         list of files to be read per replica, default is all.
         for non-compact output format, hand the folders to be read here.
     check_configs: list[list[int]]
@@ -236,6 +236,10 @@ def read_sfcf_multi(path, prefix, name_list, quarks_list=['.*'], corr_type_list=
             rep_path = path + '/' + item
             if "files" in kwargs:
                 files = kwargs.get("files")
+                if isinstance(files, list):
+                    if isinstance(files[0], list):
+                        files = files[i]
+
             else:
                 files = []
             sub_ls = _find_files(rep_path, prefix, compact, files)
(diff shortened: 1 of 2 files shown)


s-kuberski:
Hi,
regarding your comment in #225: Does your change break backward compatibility or would you consider your changes to be bug fixes?

jkuhl-uni:
With ll. 239, my changes do not break backwards compatibility.
The functionality is:
If we only have one replicum or want to read the same configs from multiple replica, we can simply hand over one list of configs. Otherwise, we are also able to hand over multiple lists, such that we can import different configs from different replica.

pyerrors/input/sfcf.py
```diff
@@ -236,6 +236,10 @@ def read_sfcf_multi(path, prefix, name_list, quarks_list=['.*'], corr_type_list=
             rep_path = path + '/' + item
             if "files" in kwargs:
                 files = kwargs.get("files")
+                if isinstance(files, list):
+                    if isinstance(files[0], list):
+                        files = files[i]
+
```
fjosw, line 242: What happens if files in to of type list[list[int]]? Maybe we should add an explicit exception here or is there a type check somewhere else that I overlooked?

pyerrors/input/sfcf.py
```diff
@@ -309,7 +313,7 @@ def read_sfcf_multi(path, prefix, name_list, quarks_list=['.*'], corr_type_list=
             w = specs[3]
             w2 = specs[4]
             if "files" in kwargs:
-                ls = kwargs.get("files")
```
fjosw, line 312: Was this a bug? Did the code work before when specifying the kwarg files?

fjosw (changes requested):
Thanks for taking care of this @jkuhl-uni. I left two small comments.

pyerrors/input/sfcf.py
jkuhl-uni: AFAIK this worked, but the thing is that the files-keyword didn't do anything in this case... With this, it would give you the option to (at least) say which ids you want, however, this is of course nto the same kind of control you have in the other cases, where you can actually say which idls you want.
For this to work, we would need to implement something similar to #185 for this read method. I coudl also take care of that, but I need some time for that I think.

pyerrors/input/sfcf.py
jkuhl-uni: Hi, sorry it took some time, I put explicit type checks in both places where `kwargs.get('files')` is used and just put two tests.

PR #232, opened 2024-04-09
[fix] Handle missing replia

s-kuberski:
Hi,

there is a bug in the error computation when one encounters the case where one combines two Obs from the same ensemble and fluctuations on one (or several) replica are not part of one of the Obses. A scaling factor needs to be applied to the fluctuation. This has been taken care of in the case of missing measurements inside a replica, but not yet on the level of the replica themselves.

I have written a first version of a fix for the case of simple Obs. If we agree that this is a good implementation or find a better one (I was not really sure when it is best to apply the scaling), this has to be extended to the arraymode.

The test routine contains the case that is erroneous in the current version and includes an analytic check of a simple case as well as some more complicated ones.

### DIFF
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -1243,6 +1243,25 @@ def derived_observable(func, data, array_mode=False, **kwargs):
         new_r_values[name] = func(tmp_values, **kwargs)
         new_idl_d[name] = _merge_idx(idl)
 
+    def _compute_scalefactor_missing_rep(obs):
+        """
+        Computes the scale factor that is to be multiplied with the deltas
+        in the case where Obs with different subsets of replica are merged.
+        Returns a dictionary with the scale factor for each Monte Carlo name.
+
+        Parameters
+        ----------
+        obs : Obs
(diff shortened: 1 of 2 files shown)


fjosw:
The fix looks good to me, I guess the scale factor could also be part of `_expand_deltas_for_merge` but I don't have a preference for the code structure. Once this is merged I would propose to release v2.11.1 such that python 3.8 user also get this fix.

s-kuberski:
Hi, 
I have adapted the pr:
- As you suggested, I have pushed the scaling in '_expand_deltas_for_merge'. This might be the safer option since it is harder to overlook the (possibly missing) scaling factor when changes are made in the future. 
- I have fixed a bug in the tests that lead to checks that did not throw an exception.
- I have extended the fix to the 'array_mode' and also added checks.

I hope that the overhead is completely negligible, especially in the array_mode, where the scaling factors are computed multiple times. I was aiming for a minimal change, but this could be refactored.

I agree that we should make a new release to distribute the bug fix.

fjosw (approved):

PR #234, opened 2024-04-14
[fix] Numpy 2 breaking changes in tests

fjosw:


s-kuberski:
Hi,
thanks for taking care of this! Will they remove 'np.NaN' in version 2.0?

PR #238, opened 2024-08-15
Multi read reading stuff multiple times

jkuhl-uni:
Hey, I was importing correlators with `read_sfcf_multi()` and found the "problem", that `bi` correlators are imported multiple times, if `len(wf2_list) > 1`. This PR corrects this and only reads the correlator once, with `wf2 = 0`.


fjosw (approved):
Looks good to me! The input/sfcf module is getting a bit complex, but I don't think it's really worth it to refactor it.

PR #241, opened 2024-09-13
fix: corrected bug that prevented combined fits with multiple x-obs in some cases

s-kuberski:
It was not possible to perform combined fits which rely on more than one x-obs when the data sets had unequal numbers of observations. The case that failed is given in the newly implemented test function. 
The bugfix ensures that the data sets are combined appropriately.

### DIFF
--- a/pyerrors/fits.py
+++ b/pyerrors/fits.py
@@ -256,7 +256,7 @@ def least_squares(x, y, func, priors=None, silent=False, **kwargs):
     if sorted(list(funcd.keys())) != key_ls:
         raise ValueError('x and func dictionaries do not contain the same keys.')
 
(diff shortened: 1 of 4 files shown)


fjosw:
We seem to have bad luck with the rng. I fixed one of the failing tests by slightly increasing the tolerance but now also `test_merge_idx` is failing. There I'm not so sure what causes this 🤔

s-kuberski:
That’s indeed weird. Everything went fine on my local machine and I would not have thought that there was randomness involved in this test.  I’ll try to resolve it.

s-kuberski:
There seemed to be a coincidence where we were (actually, it was my faulty implementation of the test) testing the equality of a range and a list in certain circumstances. Double bad luck, I would say. I've fixed the test.

fjosw (approved):

PR #245, opened 2024-11-17
[Feat] Meas class for gaussian error propagation.

fjosw:
I sometimes want to do Gaussian error propagation and find the syntax for creating a scalar cov_Obs a bit inconvenient. For this reason I propose a `Meas` class that simplifies this.
```python
import pyerrors as pe
value = 1.1
standard_error = 0.23
meas = pe.Meas(value, standard_error)
covo = pe.cov_Obs(value, standard_error ** 2, meas.names[0])
assert covo == meas
```

Standard errors do not need to be squared and if no name is provided, a uuid is generated instead.  I also propose to automatically evaluate the gamma method when an Obs exclusively consists of `cov_Obs` / `Meas` as the error estimation should be unambigous and the performance oberhead should be minimal.

In the process I also found a small bug in the cov_Obs implementation.

### DIFF
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -9,6 +9,8 @@ import matplotlib.pyplot as plt
 from scipy.stats import skew, skewtest, kurtosis, kurtosistest
 import numdifftools as nd
 from itertools import groupby
+from typing import Optional, Union
+import uuid
 from .covobs import Covobs
 
 # Improve print output of numpy.ndarrays containing Obs objects.
@@ -1354,6 +1356,9 @@ def derived_observable(func, data, array_mode=False, **kwargs):
         final_result[i_val]._value = new_val
         final_result[i_val].reweighted = reweighted
 
+        if not final_result[i_val].idl:
+            final_result[i_val].gm()
+
     if multi == 0:
         final_result = final_result.item()
 
@@ -1810,7 +1815,7 @@ def cov_Obs(means, cov, name, grad=None):
         co : Covobs
             Covobs to be embedded into the Obs
         """
-        o = Obs([], [], means=[])
+        o = Obs(samples=[], names=[], means=[])
         o._value = co.value
         o.names.append(co.name)
         o._covobs[co.name] = co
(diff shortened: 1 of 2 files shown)


pyerrors/obs.py
```diff
@@ -1822,14 +1827,51 @@ def covobs_to_obs(co):
...
+            Optional name identifier for the measurement. If none is specified, a random uuid
+            string is used instead.
+        """
+        if not isinstance(value, (float, int)):
+            raise TypeError(f"value has to be a flaot or int, not {type(value)}")
+        if not isinstance(dvalue, (float, int)):
+            raise TypeError(f"dvalue has to be a float or int, not {type(dvalue)}")
+        super().__init__(samples=[], names=[], means=[])
+        if name is None:
+            name = uuid.uuid4().hex
```
jkuhl-uni, line 1863: Hey @fjosw,
the class looks good to me. Thanks for the idea and addition. I just wanted to ask if this is something that we want. I do not mind it, and if there is a specific use case for you, we can, of course, leave it in.

fjosw:
Hi @jkuhl-uni & @s-kuberski ,

> What was the bug in the existing implementation? The missing 'float()'?

Yes, we currently compute the gradients for `__pow__` via autograd which cannot cope with `int`, so I had to add the `float` cast to make it work. We could actually provide the gradients for pow directly which should be faster. I can prepare a PR for that.

> if this is something that we want
> I ask myself if assigning automatically chosen random names to observables is something that one should do.

My underlying assumption for the random uuid is that, unlike Monte Carlo data, constants with errors are usually statistically independent of all other sources of error. If that is the case it might be easier to automatically assign a random string instead of having the user come up with unique identifiers. The option to assign a dedicated name of course still exists.

By adding the `Meas` class we could slightly increase the scope of pyerrors as a tool for Gaussian error propagation which feels a bit cumbersome with the current UI. I was considering adding another example to detail this feature but if you feel like it is too dangerous to blindly assign a random name, I'm happy to remove it or change the default.

s-kuberski:
> We could actually provide the gradients for pow directly which should be faster. I can prepare a PR for that.

That sounds good!

> By adding the Meas class we could slightly increase the scope of pyerrors as a tool for Gaussian error propagation which feels a bit cumbersome with the current UI. 

I agree that the current UI is not perfectly suited for quick error propagation, because I chose to ask for the covariance and because we require a name. If one would like to allow users to easily perform similar tasks as they would do in the 'uncertainties' package, the proposed change would be the perfect solution, I think.

> My underlying assumption for the random uuid is that, unlike Monte Carlo data, constants with errors are usually statistically independent of all other sources of error. If that is the case it might be easier to automatically assign a random string instead of having the user come up with unique identifiers. The option to assign a dedicated name of course still exists.

As soon as someone runs a script twice and saves the output somewhere to process it afterwards, I would think that a dedicated naming scheme is very important to ensure proper error estimation. If someone is not aware of this, it could easily happen that they input, say, the physical pion mass using the new class and end up with uncorrelated contributions from the same observable. I'd think that names really matter as soon as one does a proper analysis.

So probably it boils down to anticipating what users do with the package that would still provide all the means to do everything properly and the question if not asking for a name per default might make it easier to do mistakes. I'm not really sure about how to proceed.

fjosw:
> As soon as someone runs a script twice and saves the output somewhere to process it afterwards, I would think that a dedicated naming scheme is very important to ensure proper error estimation. 

If your main concern is about saving and reusing objects with automatically generated names, we could simply prevent such `Obs` to be exported. I added an explicit check to the json module which throws a `ValueError` in these cases. Do you feel like this is safe enough or are there still other silent failure modes?

s-kuberski:
Maybe we could make the name non-optional and still use an uuid if None or an empty string is passed? This would at least prompt users to think about names. Of course, it again adds a bit more luggage to the easy creation of these objects. At least, one would not have to come up with names, if one just wants to do a bit of error propagation where everything can be assumed to be uncorrelated.

jkuhl-uni:
I also thought about runtime warnings like
`Warning: No name given, using UUID`.
However, I think those would be pretty ugly and not helpful since a) the class is used as intended and a warning would still be issued and b) they could vanish in too many notifications.

PR #246, opened 2024-11-20
[Feat] Provide derivatives for pow

fjosw:
I explicitly added the derivatives for `__pow__` and `__rpow__`, previously these were estimated via autograd. On my machine this results in a ~4x speedup for pow operations, `obs ** 2` is now slightly faster than `obs * obs`.

I removed the `Obs` case from `__rpow__` as this should be covered by `__pow__` already and none of our tests fail 🤞

### DIFF
--- a/pyerrors/obs.py
+++ b/pyerrors/obs.py
@@ -861,10 +861,7 @@ class Obs:
             return derived_observable(lambda x, **kwargs: x[0] ** y, [self], man_grad=[y * self.value ** (y - 1)])
 
     def __rpow__(self, y):
-        if isinstance(y, Obs):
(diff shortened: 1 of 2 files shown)


fjosw:
We are still missing explicit derivatives for
- `abs` (I did not want to deal with the special cases)
- `arcsin`, `arccos`, `arctan`, `arcsinh`, `arccosh`, `arctanh` (probably hardly used at all)

s-kuberski:
Great, thanks! I also think that it is not necessarily worth implementing the derivatives for the remaining functions. For '__pow__' however, there might be some speedup in quite a number of real world examples.

PR #247, opened 2024-12-05
[Fix] plateaus in correlators

jkuhl-uni:
Hi,
I was tidying some of my scripts, when I found that, to me, `Corr.plateau()` and `Corr.set_prange()` have a small inconsistency:
It is possible to have a plateau over just one timeslice with `plateau()`, but `set_prange()` rejects it.
Is this intentional?
IMO, it should be possible to also set `[x,x]` as a plateau range.


s-kuberski:
I agree, your fix should make the two consistent and indeed it makes sense to allow having a plateau range of a single time slice.

fjosw (approved):

PR #248, opened 2024-12-23
[Fix] Ruff rules and more precise Excpetion types

fjosw:
This PR applies a few PEP 8 corrections and makes the types of Exception more precise.


s-kuberski (approved):
Great, thanks!

PR #249, opened 2024-12-24
[Fix] Remove mutables as default argument and fix smaller issues

fjosw:
In this PR I remove all mutables from default arguments (see for example [this article](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil)).

I also corrected smaller stilistic issues and added missing docstrings.

### DIFF
--- a/pyerrors/__init__.py
+++ b/pyerrors/__init__.py
@@ -1,4 +1,4 @@
-r'''
+r"""
 # What is pyerrors?
 `pyerrors` is a python package for error computation and propagation of Markov chain Monte Carlo data.
 It is based on the gamma method [arXiv:hep-lat/0306017](https://arxiv.org/abs/hep-lat/0306017). Some of its features are:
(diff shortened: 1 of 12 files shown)


s-kuberski:
Hi! Thanks a lot for these changes - I was not aware of the issue...
We would have to modify the call to `covariance` in the fit function. I think that we intended to have the possibility to pass `**kwargs` because it was possible to make choices regarding what method is used to compute the covariance matrix. In principle, a matrix that differs from the standard one, could now be passed directly to the routine.
One could also think to explicitly check the `**kwargs` of the fit function for entries that are relevant for the covariance matrix. With the current implementation the parameter `smooth` cannot be passed on... What do you think?

s-kuberski:
It should be sufficient to change
https://github.com/fjosw/pyerrors/blob/4089238ddde38014a40f02b2e991f0c15b1a9dc0/pyerrors/fits.py#L369
to something like 
```python
corr = covariance(y_all, correlation=True, smooth=kwargs.get('smooth', None)) 
```
to get rid of the error and to still allow to use the smoothing here. If we don't pass on the keyword, this would be a breaking change that we could at some point implement, but maybe only with a deprecation warning where we check if the keyword has been passed.

PR #251, opened 2024-12-25
[Feat] Add type hints

fjosw:
As a first step towards adding type hints, I added automatic type hints with monkeytype. Hints for the more complicated methods will need additional manual work but for the simpler methods this seems to have worked quite well.

### DIFF
--- a/pyerrors/correlators.py
+++ b/pyerrors/correlators.py
@@ -1,3 +1,4 @@
+from __future__ import annotations
 import warnings
 from itertools import permutations
 import numpy as np
@@ -9,6 +10,8 @@ from .misc import dump_object, _assert_equal_properties
 from .fits import least_squares
 from .roots import find_root
 from . import linalg
+from numpy import float64, int64, ndarray, ufunc
+from typing import Any, Callable, List, Optional, Tuple, Union
 
 
 class Corr:
@@ -42,7 +45,7 @@ class Corr:
 
     __slots__ = ["content", "N", "T", "tag", "prange"]
 
-    def __init__(self, data_input, padding=[0, 0], prange=None):
+    def __init__(self, data_input: Any, padding: List[int]=[0, 0], prange: Optional[List[int]]=None):
         """ Initialize a Corr object.
 
         Parameters
@@ -119,7 +122,7 @@ class Corr:
         self.T = len(self.content)
         self.prange = prange
 
-    def __getitem__(self, idx):
+    def __getitem__(self, idx: Union[slice, int]) -> Union[CObs, Obs, ndarray, List[ndarray]]:
         """Return the content of timeslice idx"""
         if self.content[idx] is None:
             return None
@@ -151,7 +154,7 @@ class Corr:
 
     gm = gamma_method
 
-    def projected(self, vector_l=None, vector_r=None, normalize=False):
+    def projected(self, vector_l: Optional[Union[ndarray, List[Optional[ndarray]]]]=None, vector_r: None=None, normalize: bool=False) -> "Corr":
         """We need to project the Correlator with a Vector to get a single value at each timeslice.
 
         The method can use one or two vectors.
@@ -190,7 +193,7 @@ class Corr:
             newcontent = [None if (_check_for_none(self, self.content[t]) or vector_l[t] is None or vector_r[t] is None) else np.asarray([vector_l[t].T @ self.content[t] @ vector_r[t]]) for t in range(self.T)]
         return Corr(newcontent)
 
-    def item(self, i, j):
+    def item(self, i: int, j: int) -> "Corr":
         """Picks the element [i,j] from every matrix and returns a correlator containing one Obs per timeslice.
 
         Parameters
@@ -205,7 +208,7 @@ class Corr:
         newcontent = [None if (item is None) else item[i, j] for item in self.content]
         return Corr(newcontent)
 
-    def plottable(self):
+    def plottable(self) -> Union[Tuple[List[int], List[float64], List[float64]], Tuple[List[int], List[float], List[float64]]]:
         """Outputs the correlator in a plotable format.
 
         Outputs three lists containing the timeslice index, the value on each
@@ -219,7 +222,7 @@ class Corr:
 
         return x_list, y_list, y_err_list
 
-    def symmetric(self):
+    def symmetric(self) -> "Corr":
         """ Symmetrize the correlator around x0=0."""
         if self.N != 1:
             raise ValueError('symmetric cannot be safely applied to multi-dimensional correlators.')
@@ -240,7 +243,7 @@ class Corr:
             raise ValueError("Corr could not be symmetrized: No redundant values")
         return Corr(newcontent, prange=self.prange)
 
-    def anti_symmetric(self):
+    def anti_symmetric(self) -> "Corr":
         """Anti-symmetrize the correlator around x0=0."""
         if self.N != 1:
             raise TypeError('anti_symmetric cannot be safely applied to multi-dimensional correlators.')
@@ -277,7 +280,7 @@ class Corr:
                         return False
         return True
 
-    def trace(self):
+    def trace(self) -> "Corr":
         """Calculates the per-timeslice trace of a correlator matrix."""
         if self.N == 1:
             raise ValueError("Only works for correlator matrices.")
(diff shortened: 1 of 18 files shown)


fjosw:
Not quite done but already in decent shape for a review @s-kuberski @jkuhl-uni:
- Most of the changes are just type annotations
- I changed a few lines to help mypy with understanding the type annotations
- I fixed a few smaller bugs in which explicit type checks were not consistant or did not cover certain edge cases

For now I ignored the `input/*` modules and most of `correlators.py`, the remaining modules only raise a few smaller mypy errors which I wasn't able to fix right away.

s-kuberski:
Great, thanks a lot for all of your work! I'll have a look in the upcoming week.

pyerrors/correlators.py
```diff
@@ -42,7 +45,7 @@ class Corr:
 
     __slots__ = ["content", "N", "T", "tag", "prange"]
 
-    def __init__(self, data_input, padding=[0, 0], prange=None):
+    def __init__(self, data_input: Any, padding: list[int]=[0, 0], prange: Optional[list[int]]=None):
```
jkuhl-uni, line 48: SHouldn't this be `list[Obs]`? Alternatively, this could be something Obs-like, I suppose, such as CObs?

jkuhl-uni:
Hey @fjosw,
thanks for all the work, I am looking through it right now, it looks excellent to me. I hope it was okay that I pushed a few small changes. If not, feel free to revert them, as you please :)

pyerrors/correlators.py
```diff
@@ -277,19 +281,19 @@ def is_matrix_symmetric(self):
                         return False
         return True
 
-    def trace(self):
+    def trace(self) -> "Corr":
         """Calculates the per-timeslice trace of a correlator matrix."""
         if self.N == 1:
             raise ValueError("Only works for correlator matrices.")
-        newcontent = []
+        newcontent: list[Union[None, float]] = []
```
s-kuberski, line 288: Similar to Justus' question above: `newcontent` should be a `list of `None` or `Obs` (or `CObs`?), right?

pyerrors/correlators.py
```diff
@@ -701,7 +698,7 @@ def second_deriv(self, variant="symmetric"):
         else:
             raise ValueError("Unknown variant.")
 
-    def m_eff(self, variant='log', guess=1.0):
+    def m_eff(self, variant: str='log', guess: float=1.0) -> "Corr":
```
s-kuberski, line 701: A very naive question and probably that is just something that was automatically done: When shoult it be `Optional[str]` (as above) and when `str` (as here)?

fjosw: I think it should be `Optional[str]` when the value can also be `None`, I think in this case one really needs to pass a string, right?

pyerrors/correlators.py
```diff
@@ -868,7 +865,7 @@ def set_prange(self, prange):
         self.prange = prange
         return
 
-    def show(self, x_range=None, comp=None, y_range=None, logscale=False, plateau=None, fit_res=None, fit_key=None, ylabel=None, save=None, auto_gamma=False, hide_sigma=None, references=None, title=None):
+    def show(self, x_range: Optional[list[int]]=None, comp: Optional[Corr]=None, y_range: None=None, logscale: bool=False, plateau: None=None, fit_res: Optional[Fit_result]=None, fit_key: Optional[str]=None, ylabel: None=None, save: None=None, auto_gamma: bool=False, hide_sigma: None=None, references: None=None, title: None=None):
```
s-kuberski, line 868: The types of `y_range`, `plateau`, `ylabel`, `save`, `hide_sigma`, `references`, `title` have not been recognized.

pyerrors/correlators.py
fjosw: Optional[X] is equivalent to X | None (or Union[X, None]).

https://docs.python.org/3/library/typing.html#typing.Optional

pyerrors/correlators.py
```diff
@@ -1084,18 +1081,18 @@ def __str__(self):
 
     __array_priority__ = 10000
 
-    def __eq__(self, y):
+    def __eq__(self, y: Union[Corr, Obs, int]) -> ndarray:
```
s-kuberski, line 1084: I guess `y` can be `Any`?

pyerrors/correlators.py
```diff
@@ -1084,18 +1081,18 @@ def __str__(self):
...
         if isinstance(y, Corr):
             comp = np.asarray(y.content, dtype=object)
         else:
             comp = np.asarray(y)
         return np.asarray(self.content, dtype=object) == comp
 
-    def __add__(self, y):
+    def __add__(self, y: Any) -> "Corr":
```
s-kuberski, line 1091: In contrast, here, `y` should be one of the types that don't raise an exception.

pyerrors/correlators.py
```diff
@@ -1190,11 +1187,11 @@ def __rmatmul__(self, y):
         else:
             return NotImplemented
 
-    def __truediv__(self, y):
+    def __truediv__(self, y: Union[Corr, float, ndarray, int]) -> "Corr":
```
s-kuberski, line 1190: `Obs` and `CObs` are missing here. BTW: does that just mean that we don't call the function with these types in our tests?

pyerrors/correlators.py
fjosw: When a type has not been recognised, this indicates that we don't have proper test coverage for this call 😬

pyerrors/correlators.py
fjosw: Yes, that's my understanding. monkeytype derived these type hints from our tests.

pyerrors/correlators.py
```diff
@@ -1244,37 +1241,37 @@ def __truediv__(self, y):
         else:
             raise TypeError('Corr / wrong type')
 
-    def __neg__(self):
+    def __neg__(self) -> "Corr":
         newcontent = [None if _check_for_none(self, item) else -1. * item for item in self.content]
         return Corr(newcontent, prange=self.prange)
 
-    def __sub__(self, y):
+    def __sub__(self, y: Union[Corr, float, ndarray, int]) -> "Corr":
```
s-kuberski, line 1248: incomplete list for `y`.

s-kuberski: Okay, I stop here for the rest of the file. There are a few more instances below, where we have to adjust `y.

pyerrors/correlators.py
```diff
@@ -1356,7 +1353,7 @@ def return_imag(obs_OR_cobs):
 
         return self._apply_func_to_corr(return_imag)
 
-    def prune(self, Ntrunc, tproj=3, t0proj=2, basematrix=None):
+    def prune(self, Ntrunc: int, tproj: int=3, t0proj: int=2, basematrix: None=None) -> "Corr":
```
s-kuberski, line 1356: `basematrix` should be an `Optional[ndarray]`.

s-kuberski:
Thansk again for starting this. I'm still in `correlators.py` and have discovered quite a few places, where I think that the type hints should be changed, if one wants them to be correct. What's our strategy? Shall I continue to point them out and you change them if you agree? Or shall I directly adapt? I guess it's not breaking anything if some of the hints are incorrect, but if we introduce them in the first place, I think we should make an effort to have them correct and hopefully complete.

fjosw:
Feel free to get started with fixing stuff, you should have permissions to push to the branch, I think I won't have time to work on this in the comming days. My strategy so far was to run `mypy` to point out typing issues and go from there. As @jkuhl-uni suggested it might be worth defining a Obs-like type which can then be used for the overloaded methods.

s-kuberski:
I've tried to go through the changes and to address the issues that I raise above and a few more things.

fjosw:
I resolved the merge conflicts and fixed a few smaller issues but now one of the tests `test_rwms` is failing. Can you have a look @jkuhl-uni ?

jkuhl-uni:
I had a look and found a solution for now, there are a few minor things to figure out. I hope to get it done today :)

jkuhl-uni:
... well that was easy.
I'd like to disentangle some of the openQCD tests (in `test_rwms` we are testing the `read_rwms`, `extract_t0` and `extract_w0` methods), but I'll do that in a different PR.

jkuhl-uni:
Hi,
I looked through the code again and made mainly small changes, as to improve some type hints.
The PR looks good to me otherwise.

pyerrors/input/hadrons.py
```diff
@@ -382,7 +384,7 @@ def __array_finalize__(self, obj):
         self.mom_out = getattr(obj, 'mom_out', None)
 
 
-def read_ExternalLeg_hd5(path, filestem, ens_id, idl=None):
+def read_ExternalLeg_hd5(path: str, filestem: str, ens_id: str, idl: Optional[range]=None):
```
jkuhl-uni:
```suggestion
def read_ExternalLeg_hd5(path: str, filestem: str, ens_id: str, idl: Optional[Union[list, range]]=None):
```

Issue #256, opened 2025-02-11
Suspicious dvalues for mean of multiple ensembles


JanNeuendorf:
### Describe the issue:

When we have an observable that is calculated on multiple ensembles (with similar results and errors) I would expect the the error of the whole mean to decrease with the number of independent ensembles added. This does not seem to be the case. 
Is there a reason, why the error should increase?

### Code example:

```python
def make_test_obs(number_of_ensembles:int):
    samples=[[1.1,0.9,0.9,1.0,1.0,1.1,1.1,0.9]]*number_of_ensembles
    names=[f"ensemble_{i}" for i in range(number_of_ensembles)]
    o=pe.Obs(samples,names)
    o.gm()
    return o
lsp=[1,10,100,500,1000]
plt.plot(lsp,[make_test_obs(n).dvalue for n in lsp],"bo"),
plt.show()
```

### Error message:

```shell

```

### Runtime information:

system    	 Linux
python    	 3.12.7
pyerrors  	 2.13.0
numpy     	 2.2.2
scipy     	 1.15.1
matplotlib	 3.10.0
pandas    	 2.2.3

s-kuberski:
To summarize, in your example, you are adding an independent contribution to the error with each additional ensemble. The resulting error of your observable should thus increase as $\sqrt{N}$ because you have $\sigma = \sqrt{\sigma_0^2 + \sigma_1^2 + \dots + \sigma_{N-1}^2}$.

JanNeuendorf:
Yes, I did not use replicas on purpose and I agree that it appears to work correctly for them. 
I do not get, why I would expect it to increase with $\sqrt{N}$.
If I set up $N$ ensembles and perform $N$ measurements with similar errors, I would expect $\frac{1}{\sqrt{N}}$ because I have independent samples.  You can argue for this in the following way:
Lets say that my individual ensembles all have a $\sigma$ of around 1 (and they actually come from uncorrelated samples with a nice Gaussian distribution).
So when I ask the program for their individual dvalue, I get a meaningful answer. 
"You will find 68% of values in this range around the mean".
When I now combine 100 of those ensemble, the dvalue would gain a factor of 10. 
So what does this error mean?

JanNeuendorf:
> So I guess the question is if it should be allowed to construct the Obs in the way you do it in your example, right?

I guess so. But I can construct it this way and the mean value does correspond to the average, so why would I expect the errors to work different from an average?

JanNeuendorf:
[...]

PR #258, opened 2025-02-17
[Fix] Removed the possibility to create an Obs from data on several replica

s-kuberski:
This pull request addresses #256:

It removes the possibility to create an `Obs` from data that is based on several ensembles. Whereas the feature is needed internally, it can be considered dangerous for users as the behavior of the function might be unexpected. To allow for internal creation of such `Obs` in intermediate steps, the creation is still allowed if `means` are passed to the function.

Further changes that I introduced along the way because they are coupled to the same problem. It is possible that I have overlooked something. In this case, these changes can be discussed and reverted.
- It is not possible anymore to reweight an `Obs` that already contains several ensembles. The only setting where I see that this could be a valid procedure is the case where algorithmic parameters that do not affect the reweighting but only the autocorrelation time have been changed such that two independent streams are not real replica and instead treated as separate ensembles. Also in this case, I would propose to force the user to reweight before combining the observables.
- It is not possible anymore to correlate two observables that are based on several ensembles. I think that this is a feature that has to be applied to primary observables instead of derived observables, because otherwise averages are not properly defined.
- The function `merge_obs` has been adapted to the same rules. The function header is extended to better explain what is done here.

Apart form this, the biggest changes in the code are in the JSON and `dobs` I/O, where I had to adapt to the new way to create the `Obs` and in the tests, where we had used the illegal creation in some cases.

It would be an option to add a routine that combines observables from several pseudo-replica by averaging them, either weighted by the number of configs or by the number of independent configs.

### DIFF
--- a/pyerrors/input/dobs.py
+++ b/pyerrors/input/dobs.py
@@ -529,7 +529,8 @@ def import_dobs_string(content, full_output=False, separator_insertion=True):
                 deltas.append(repdeltas)
                 idl.append(repidl)
 
-        res.append(Obs(deltas, obs_names, idl=idl))
+        obsmeans = [np.average(deltas[j]) for j in range(len(deltas))]
+        res.append(Obs([np.array(deltas[j]) - obsmeans[j] for j in range(len(obsmeans))], obs_names, idl=idl, means=obsmeans))
         res[-1]._value = mean[i]
     _check(len(e_names) == ne)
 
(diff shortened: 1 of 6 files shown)


fjosw (approved):
Looks good to me!

fjosw:
Maybe @JanNeuendorf can confirm that this actually addresses the issue her raised.

JanNeuendorf:
I think that it solves the issue because I see no remaining way to get a multi-ensemble Obs without taking an explicit average.

fjosw (approved):

PR #261, opened 2025-04-17
Fixed index in GEVP example

JanNeuendorf:
I think the description does not match the cell above.


s-kuberski:
It seems as I'd have to make adjustments in `json.py` to make the linter happy. I'll look at it.

fjosw (approved):

PR #270, opened 2025-10-10
[ci] Add python 3.14 runners for pytest workflow

fjosw:


fjosw:
Python 3.14 runners were failing with the `-Werror` option because we have a new SyntaxWarning in autograd
```
../../../../../opt/hostedtoolcache/Python/3.14.0/x64/lib/python3.14/site-packages/autograd/wrap_util.py:38
  /opt/hostedtoolcache/Python/3.14.0/x64/lib/python3.14/site-packages/autograd/wrap_util.py:38: SyntaxWarning: 'return' in a 'finally' block
    return f
 ```
 
 I'll try to resolve that in autograd.

fjosw:
See https://github.com/HIPS/autograd/issues/734

PR #274, opened 2025-10-20
[Fix] corrected expected_chisquare by adding the number of priors

PiaLJP:
The number of priors needs to be included in the expected chisquare, see eq. (D.9) in section "D.2 Priors and the Bayesian approach" in arXiv:2209.14188v2 ("On ﬁts to correlated and auto-correlated data").
Note that the expected chisquare is now correct in the case of priors that are not correlated with the data fitted. 
For correlated priors further alterations would be needed. (The covariance (from eq. (D.9)) is no longer be diagonal w.r.t the priors so you cannot just add the number of priors.)

### DIFF
--- a/tests/fits_test.py
+++ b/tests/fits_test.py
@@ -1611,3 +1611,81 @@ def old_prior_fit(x, y, func, priors, silent=False, **kwargs):
         qqplot(x, y, func, result)
 
     return output
+
+def test_dof_prior_fit():
+    """Performs an uncorrelated fit with a prior to uncorrelated data then
+    the expected chisquare and the usual dof need to agree"""
+    N = 5
+
+    def fitf(a, x):
+        return a[0] + 0 * x
+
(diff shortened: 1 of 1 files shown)


s-kuberski:
Hi Pia,

thanks a lot for adding this, it has been on my list for quite some time now. 

I agree that, the code will still not be correct in the case where the prior is correlated with the data. This should be addressed at some point. I had started to implement the proper solution, but then I was hesitant because, in principle, also the notion of a correlated fit would then have to be changed to a covariance matrix that also covers the contribution of the priors, so both cases should be handled together.

For now, it is good to fix the obvious bug for the standard case of an external prior.

s-kuberski:
Sorry, I closed this by mistake. Before merging it, would it make sense to write a small test that ensures that`chisquare_by_dof` for an uncorrelated fit to uncorrelated data, including priors, matches `chisquare_by_expected_chisquare`?

PR #275, opened 2025-10-30
Feat/idl func

jkuhl-uni:
This PR adds the option for the user to supply their own function to extract the number of the configuration from file names (or in the append mode the line that includes the gauge name).

This is particularly useful for the append-mode, as one may have the case that the configuration file has a postfix (for example `.lex`).

It is a fix to a problem I was made aware of via PM.

While the fix introduces yet another option for the SFCF read-in, this was a way to keep things backwards compatible.


s-kuberski:
Hi,

adding an optional argument to the I/O function is probably not an issue if it provides a functionality that is useful. Could you implement tests for the new feature (and fix the linting errors)?

fjosw (approved):

PR #282, opened 2026-04-20
[chore] Stricter ruff rules

fjosw:
This PR imposes stricter ruff rules and fixes the parts which raise errors now. Main changes:
- Remove mutables as default argument (was part of https://github.com/fjosw/pyerrors/pull/249)
- Removes printf style string formatting
- Imposes strict zipping


jkuhl-uni (commented):
lgtm

jkuhl-uni:
Thank you for all the work on the code style!
I think having these slightly stricter rules makes sense. I have to familiarize myself with some of the reasoning behind the ruff rules, but I think it is a good idea overall.

s-kuberski:
Hi! Thanks for taking the time to implement this! I am still in the process of looking through the changes. 

I was wondering about the change of the handling of random numbers. I understand that switching to `np.random.default_rng()` is considered to be the proper modern solution. However, I am asking myself if this change, hidden inside the large number of other changes, could lead to unexpected behavior. 

As I understand it, changing the numpy random seed will not affect the state of the random number generator anymore, right? So there won't be any way to fix the seed in the random number generator in the future, when its initialization is hidden in the module. Could this lead to problems?

fjosw:
After quite some time I had another look at this PR. I agree with leaving the RNG changes out of this PR and potentially do them separately or not at all. I removed the rng changes from the diff. Could you review the changes that are left @s-kuberski @jkuhl-uni ?

jkuhl-uni (approved):

jkuhl-uni:
Hi,
thank you again for the work.
I think moving the changes to another PR is a good idea. I haven't had time to think more about the calls within NumPy. If the current call from pyerrors is considered legacy, I hope there is an alternative that would have the same properties in regard to the seed and instantiation of the RNG as the legacy call.

s-kuberski:
Thanks for the adaption! This looks good to me and can be merged from my side.

There are only two very small things we might still want to fix before merging, that I noticed in lines that were touched in this pull request:

* in pyerrors/linalg.py, in prune, the error message for Ntrunc >= basematrix.N seems to have the inequality reversed/misleading;
* in pyerrors/covobs.py, the error message for a non-symmetric covariance matrix is missing a closing parenthesis.

Neither is blocking from my side, these are really minor.

PR #288, opened 2026-07-15
[Fix] Fix off-by-one in Covobs pos bounds check

fjosw:
Follow up to the making exception messages more explicit: This highlighted a small bug that let to a misleading exception message.


s-kuberski (approved):
Very good, thanks for spotting it!

